PR #48, opened 2026-08-02
Fix #47: keep test-compile for skipped modules whose test-jars are consumed in-reactor

### LINKED ISSUE #47: skip-tests mode can produce empty/missing test-jars for modules that are themselves intra-reactor test-jar dependencies
### Summary

In `scalpel.mode=skip-tests`, `ScalpelLifecycleParticipant.applySkipTests()` sets
`maven.test.skip=true` on modules it decides don't need testing (upstream-only modules when
`skipTestsForUpstream=true`, excluded downstream modules, or modules unrelated to the affected
set). `maven.test.skip` is honored by the Compiler Plugin as well as Surefire/Failsafe, so it
also suppresses `test-compile`, not just test execution. If the skipped module produces a
`test-jar` (via `maven-jar-plugin:test-jar`) that another reactor module consumes intra-reactor
via `<type>test-jar</type>`, that consumer's test-compile fails because the producer's
`target/test-classes` was never populated.

`ReactorTrimmer` already has a test-jar-aware check
(`hasTestJarDependency(downstream, upstream)`), but it's only used to decide whether a
**downstream** module should be pulled into the build set when an **upstream** module had
test-only source changes. There's no equivalent check in `applySkipTests()` for the reverse
direction: before applying `maven.test.skip=true` to a module, nothing checks whether some
other module still slated to run tests depends on that module's test-jar.

### Reproduction (real-world)

Apicurio Registry ran a Scalpel shadow-group evaluation
(`-Dscalpel.mode=skip-tests -Dscalpel.skipTestsForUpstream=true`) against its live reactor
(`-pl app -am`, 0.3.9). Across 121 control-vs-shadow shard comparisons over ~18 days, 56 diverged
(Scalpel-side failure, control passing); a sample traced 91% of those to this exact mechanism:
`schema-util-common` (a module producing a test-jar consumed by `schema-util/avro`,
`schema-util/iceberg`, `schema-util/json`) was repeatedly marked upstream-only and skip-tested,
and its 3 test-jar consumers failed test-compile simultaneously every time it happened.

### Minimal repro shape

```
reactor/
|-- module-common/        (produces test-jar via maven-jar-plugin:test-jar)
|-- module-a/              (depends on module-common test-jar, <scope>test</scope>)
`-- module-b/              (source change here; depends on module-common main jar only)
```
With `-Dscalpel.mode=skip-tests -Dscalpel.skipTestsForUpstream=true`, a change to `module-b`
marks `module-common` as upstream-only, so `maven.test.skip=true` gets set, so `module-a`'s
test-compile fails looking for `module-common`'s test classes.

### Suggested fix

Before `applySkipTests()` sets `maven.test.skip=true` on a candidate module, check whether any
module still in the "will be tested" set has a `<type>test-jar</type>` dependency on it
(reusing/extending `ReactorTrimmer.hasTestJarDependency`, generalized across the full build set
rather than only immediate upstream/downstream pairs). If so, exclude it from the skip set.

### Scope note

None of the existing `skip-tests-upstream` / `skip-tests-downstream-exclusion` /
`skip-tests-downstream-exclusion-report` integration tests exercise test-jar packaging or
dependencies; this interaction isn't covered by the current IT suite. Happy to contribute a
minimal IT scenario alongside a fix if useful.

Scalpel version: 0.3.9. Registry's own shadow-group evaluation write-up (for context, not a
request to read the whole thing): `Apicurio/apicurio-registry` discussion #8364.

paoloantinori:
Fixes #47.

## Mechanism

`applySkipTests()` implements skip decisions with `maven.test.skip=true`, which the Compiler Plugin also honors, so a skipped module's test-compile never runs and its `maven-jar-plugin:test-jar` artifact is never produced. Any module still running tests that consumes that test-jar in-reactor (`<type>test-jar</type>`) then fails dependency resolution. In the real-world case that motivated #47 (Apicurio Registry's shadow evaluation), one skipped producer failed all shards at once, accounting for 91% of the observed Scalpel-side failures.

## Remedy

A fix-up pass (`softenTestJarProducers`) runs after the skip partition: for skip candidates whose test-jar is consumed by any module that will compile tests, the skip is softened from `maven.test.skip=true` to `skipTests=true`: test-compile and test-jar packaging still happen, only Surefire/Failsafe execution is skipped, so the optimization is preserved. The pass iterates to a fixpoint so chained producers (A's test-jar consumed by B's tests, B's by a test-runner) are handled, and it considers skip candidates from both populations (trim build set and outside it). `ReactorTrimmer.hasTestJarDependency` is reused (promoted to package-private) rather than duplicating the check.

Softening deliberately runs after `applyPerCategoryArgs` so user-configured `upstreamArgs`/`downstreamArgs` cannot reintroduce the failure on a consumed producer (comment in code states the invariant). The summary log now reports the third bucket: `Testing X, N compile-only (test-jar producers), skipping tests on Z of Y`.

## Tests

- New IT `skip-tests-test-jar` (module-common producing a test-jar, module-a consuming it with a real JUnit test, module-b as the changed module): written red-first against the unpatched code, where it reproduces the exact production failure (`Could not find artifact ...module-common:jar:tests`); green with the fix, asserting build success, producer executions skipped, consumer tests ran.
- Unit tests including two fixpoint guards: a 3-module chain ordered so only a second iteration can soften the first producer (verified to fail under a single-pass regression), and a producer outside the trim build set.
- Full `./mvnw clean verify -e -B -P run-its`: 23/23 ITs, extension3 135/135 unit tests, BUILD SUCCESS.

## Known inherited limitation (out of scope)

`hasTestJarDependency` matches the canonical `<type>test-jar</type>` form only, not the equivalent `<classifier>tests</classifier><type>jar</type>` spelling: pre-existing behavior, unchanged here, noted in the commit body.

## Summary by CodeRabbit

* **New Features**
  * Improved skip-tests mode for projects that consume test JARs.
  * Producer modules remain available for test compilation while test execution stays skipped.
  * Added support for chained and out-of-scope test JAR dependencies.

* **Tests**
  * Added unit and integration coverage for test JAR creation, dependency propagation, reactor completion, and downstream test execution.

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
@@ -53,6 +53,7 @@ import org.slf4j.LoggerFactory;
 class ScalpelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
 
     private static final String MAVEN_TEST_SKIP = "maven.test.skip";
+    private static final String SKIP_TESTS = "skipTests";
     private static final String GLOB_PREFIX = "glob:";
 
     private final Logger logger = LoggerFactory.getLogger(getClass());
@@ -600,12 +601,69 @@ class ScalpelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
         // Apply per-category args
         applyPerCategoryArgs(trimResult, config);
 
+        // Softening must have the last word on maven.test.skip/skipTests: it runs after
+        // applyPerCategoryArgs so user-configured upstreamArgs/downstreamArgs cannot
+        // reintroduce the #47 failure on a consumed test-jar producer.
+        Set<MavenProject> softenedProjects = softenTestJarProducers(testProjects, skippedProjects);
+
         logger.info(
-                "Scalpel: Testing {} of {} modules, skipping tests on {} modules: {}",
+                "Scalpel: Testing {}, {} compile-only (test-jar producers), skipping tests on {} of {} modules: {}",
                 testProjects.size(),
-                allProjects.size(),
+                softenedProjects.size(),
                 skippedProjects.size(),
+                allProjects.size(),
                 keys(skippedProjects));
+    }
+
+    /**
+     * Modules slated for a full test skip (maven.test.skip=true) also skip test-compile, which
+     * suppresses their test-jar. If some other module that will still run tests (directly, or
+     * because it was already softened by a previous pass) depends on that test-jar
+     * (&lt;type&gt;test-jar&lt;/type&gt;), its test-compile would fail looking for classes that
+     * were never built. Soften those producers: drop maven.test.skip and use skipTests=true
+     * instead, which only disables the surefire/failsafe execution while leaving test-compile
+     * (and jar:test-jar) intact. Softening a producer can itself depend on another skipped
+     * producer's test-jar (chained test-jar consumers), so this iterates to a fixpoint.
+     */
+    private Set<MavenProject> softenTestJarProducers(
+            List<MavenProject> testProjects, List<MavenProject> skippedProjects) {
+        Set<MavenProject> compilingTests = new LinkedHashSet<>(testProjects);
+        Set<MavenProject> softenedProjects = new LinkedHashSet<>();
+        boolean changed = true;
+        while (changed) {
+            changed = false;
+            for (MavenProject candidate : skippedProjects) {
+                if (softenedProjects.contains(candidate)) {
+                    continue;
+                }
+                for (MavenProject consumer : compilingTests) {
+                    if (reactorTrimmer.hasTestJarDependency(consumer, candidate)) {
+                        candidate.getProperties().remove(MAVEN_TEST_SKIP);
+                        candidate.getProperties().setProperty(SKIP_TESTS, "true");
+                        softenedProjects.add(candidate);
+                        compilingTests.add(candidate);
+                        if (logger.isDebugEnabled()) {
+                            logger.debug(
+                                    "Scalpel: Keeping test-compile for {} because its test-jar is consumed"
+                                            + " in-reactor by {} (softened: skipTests=true instead of"
+                                            + " maven.test.skip=true)",
+                                    key(candidate),
+                                    key(consumer));
+                        }
+                        changed = true;
+                        break;
+                    }
+                }
+            }
+        }
+        skippedProjects.removeAll(softenedProjects);
+        if (!softenedProjects.isEmpty()) {
+            logger.info(
+                    "Scalpel: {} modules had test-compile restored for in-reactor test-jar consumers: {}",
+                    softenedProjects.size(),
+                    keys(softenedProjects));
+        }
+        return softenedProjects;
     }
 
     private boolean usesChangedPlugin(MavenProject project, Set<String> changedPluginGAs) {
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
@@ -636,13 +636,14 @@ class ScalpelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
                 if (softenedProjects.contains(candidate)) {
                     continue;
                 }
-                for (MavenProject consumer : compilingTests) {
+                // Iterate a snapshot: compilingTests grows in the loop body as softened producers
+                // become potential consumers for later candidates, and mutating the live set during
+                // iteration would throw ConcurrentModificationException.
+                for (MavenProject consumer : new ArrayList<>(compilingTests)) {
                     if (reactorTrimmer.hasTestJarDependency(consumer, candidate)) {
                         candidate.getProperties().remove(MAVEN_TEST_SKIP);
                         candidate.getProperties().setProperty(SKIP_TESTS, "true");
                         softenedProjects.add(candidate);
-                        // Safe despite iterating compilingTests: the break below exits before the
-                        // iterator can advance, so no ConcurrentModificationException.
                         compilingTests.add(candidate);
                         if (logger.isDebugEnabled()) {
                             logger.debug(
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
@@ -641,6 +641,8 @@ class ScalpelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
                         candidate.getProperties().remove(MAVEN_TEST_SKIP);
                         candidate.getProperties().setProperty(SKIP_TESTS, "true");
                         softenedProjects.add(candidate);
+                        // Safe despite iterating compilingTests: the break below exits before the
+                        // iterator can advance, so no ConcurrentModificationException.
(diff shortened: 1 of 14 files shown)


extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -600,14 +601,71 @@ && hasChangedTransitiveDependency(project, session, changedManagedDepGAs, resolv
...
+    /**
+     * Modules slated for a full test skip (maven.test.skip=true) also skip test-compile, which
+     * suppresses their test-jar. If some other module that will still run tests (directly, or
+     * because it was already softened by a previous pass) depends on that test-jar
+     * (&lt;type&gt;test-jar&lt;/type&gt;), its test-compile would fail looking for classes that
+     * were never built. Soften those producers: drop maven.test.skip and use skipTests=true
+     * instead, which only disables the surefire/failsafe execution while leaving test-compile
+     * (and jar:test-jar) intact. Softening a producer can itself depend on another skipped
+     * producer's test-jar (chained test-jar consumers), so this iterates to a fixpoint.
+     */
```
gnodet, line 627: **Suggestion (non-blocking):** `compilingTests.add(candidate)` modifies the `LinkedHashSet` while the inner for-each iterator is open. The `break` on line 638 saves this from a `ConcurrentModificationException`, but the invariant is fragile — a future maintainer removing the `break` would silently introduce a CME. Consider adding a brief comment:

```suggestion
                    softenedProjects.add(candidate);
                    compilingTests.add(candidate); // break below exits before iterator advances — no CME
```

it/extension3-its/src/it/skip-tests-test-jar/verify.groovy
```diff
@@ -0,0 +1,58 @@
...
+def softenedLine = lines.find { it.contains('test-compile restored') && it.contains('module-common') }
+assert softenedLine != null : "Expected the aggregate summary to report module-common's test-compile was restored (softened), got:\n$log"
+
+// Returns the log lines belonging to the plugin execution section starting at sectionStart
+// (a "--- goal ... ---" header line), up to (excluding) the next such header. Plugin artifact
+// download output can interleave arbitrarily many lines within a section, so bound by the next
+// section header rather than a fixed line count.
+def sectionText = { int sectionStart ->
+    int nextSection = (sectionStart + 1..<lines.size()).find { i -> lines[i].contains('[INFO] --- ') } ?: lines.size() - 1
+    lines[sectionStart..nextSection].join('\n')
```
gnodet: **Suggestion (non-blocking):** The comment on line 35 says "up to (excluding) the next such header", but Groovy's `..` range is inclusive on both ends, so `nextSection` (the next header line) is included. Practically harmless since the assertions search for strings that wouldn't appear in a Maven plugin header, but `lines[sectionStart..<nextSection]` would match the stated intent.

```suggestion
    lines[sectionStart..<nextSection].join('\n')
```

gnodet (approved):
## Review Summary

All review questions resolved — no blocking issues found. This is a well-crafted fix with thorough test coverage.

The approach (soften to `skipTests=true` rather than un-skipping entirely) preserves the optimization while fixing the test-jar dependency resolution failure. The fixpoint iteration handles chained test-jar producers correctly with convergence guarantees.

### Findings

**1. Suggestion (non-blocking): inline comment on `break`-after-`add` invariant**

In `softenTestJarProducers()`, `compilingTests.add(candidate)` modifies the `LinkedHashSet` while the inner for-each iterator is technically open. The immediately following `break` saves this from a `ConcurrentModificationException`, but a future maintainer removing the `break` would silently introduce one. A brief comment would protect this invariant:

```java
compilingTests.add(candidate);
// break below exits before the iterator advances — no ConcurrentModificationException
```

**2. Suggestion (non-blocking): verify.groovy `sectionText` inclusive range**

The comment on line 35 says "up to (excluding) the next such header", but `lines[sectionStart..nextSection]` uses Groovy's inclusive `..` range, so the next header line is included in the joined text. Practically harmless (the assertions search for strings that wouldn't appear in a `[INFO] --- plugin:goal ---` header), but `lines[sectionStart..<nextSection]` would match the stated intent.

### Project Standards

| Rule | Status |
|------|--------|
| Commit format | ✅ `Fix #47: ...` |
| Branch naming | ✅ `fix/skip-tests-test-jar-consumers` |
| No new production dependencies | ✅ |
| Tests present | ✅ 3 unit tests + 1 IT |
| Known limitation documented | ✅ `<classifier>tests</classifier>` form noted |

This review does not replace specialized tools such as CodeRabbit, Sourcery, or SonarCloud.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
paoloantinori: Done in fd927f0. Added a comment on the `compilingTests.add(candidate)` line explaining that the `break` below exits before the iterator advances, so the in-flight modification is CME-safe. Reworded slightly from your suggestion.

it/extension3-its/src/it/skip-tests-test-jar/verify.groovy
paoloantinori: Done in fd927f0, with one extra change beyond the literal suggestion.

Your `..<` fix is correct on the found-header path: the next `---` header is now excluded, matching the comment's "excluding" wording. Applied alone, though, it would truncate the last section: the no-next-header fallback on the line above is `?: lines.size() - 1`, and an exclusive end of `lines.size() - 1` drops the final log line. So I also widened the fallback to `?: lines.size()`. With `(..<)` plus `lines.size()`, the last section keeps all its lines and every other section excludes its trailing header. The two assertions here read non-final sections, so they pass either way, but the helper is now correct for any section.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
paoloantinori: Took this one step further in 65dbae2 rather than leaving the comment.

The documenting comment made the safety depend on the `break` staying in place, so I replaced it with a structural fix: the inner loop now iterates a snapshot, `new ArrayList<>(compilingTests)`, while the load-bearing `compilingTests.add(candidate)` still mutates the live set. Mutating a set while iterating its copy is safe regardless of the `break`, so the `break` is no longer a correctness hinge and a future maintainer can remove it without introducing a CME. The `add` itself is kept because it is load-bearing for within-pass chaining (a softened producer becomes a consumer for later candidates in the same pass); the snapshot is taken per candidate, so it still sees producers softened earlier in the pass. All 53 lifecycle-participant tests still pass, including the chained-producer fixpoint test.

gnodet (approved):
## Review: APPROVE ✅

All factual claims verified against the source code.

### What was checked

| Area | Verdict |
|---|---|
| Fixpoint loop termination and correctness | ✅ Guaranteed: bounded by `skippedProjects.size()`, each pass softens ≥1 new candidate. `ConcurrentModificationException` prevented via `new ArrayList<>` snapshot. |
| `maven.test.skip` → `skipTests` softening | ✅ Correct Maven semantics: `skipTests=true` preserves test-compile and test-jar while skipping surefire/failsafe. |
| Ordering invariant (after `applyPerCategoryArgs`) | ✅ Softening runs last so user-configured args cannot reintroduce the failure. |
| `hasTestJarDependency` visibility change | ✅ Minimal: `private` → package-private, method body untouched, Javadoc added. |
| Test coverage | ✅ Excellent: 3 unit tests (basic, chained fixpoint, outside-build-set) + 1 end-to-end IT with red-first methodology. |
| Known limitation (`<classifier>tests</classifier>` spelling) | ✅ Pre-existing, honestly documented, not introduced by this PR. |

High-quality bug fix with thorough testing and clear invariant documentation.

PR #60, opened 2026-08-08
chore: align report schema with forward-compatibility policy

gnodet:
## Summary

Fixes two issues found during review of PR #53 (report schema v2):

- **Remove `additionalProperties: false`** from both the root object and the `affectedModule` definition. The README (lines 167-169) states that new optional fields may be added within a major version and consumers should ignore unknown fields. Setting `additionalProperties: false` contradicts this — any future optional field would break validators without a schema version bump.

- **Remove `EXCLUDED_DOWNSTREAM` from the `reasons` enum.** This value is never used as a module-inclusion reason in the codebase. It is only assigned to `testsSkippedReason` (see `ScalpelLifecycleParticipant.java`). Including it in the `reasons` enum is misleading. The `testsSkippedReason` field description still references `EXCLUDED_DOWNSTREAM`, which is correct.

## Test plan

- [x] Full `./mvnw clean install -B` passes with no test failures

---

Review finding from PR #53.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,6 @@ consumers may need to handle.
 | `DOWNSTREAM_DEPENDENT` | Included as a downstream dependent (via `alsoMakeDependents`) |
 | `DOWNSTREAM_TEST` | Included as a downstream dependent via test-scoped dependency only |
 | `FORCE_BUILD` | This module was force-included via `forceBuildModules` |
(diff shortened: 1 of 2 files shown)


gnodet:
@coderabbitai review

PR #66, opened 2026-08-17
fix: stop mapping repo-root files to the root aggregator module

paoloantinori:
Repo-root files (README.md, .gitignore, LICENSE) currently map to the root aggregator module, so touching them marks the root POM affected and pulls the whole reactor into the build. A repo-root file is not a change to any module.

Extracts an `isFileInProjectSubtree` helper and uses it so only files inside a module's subtree map to that module. Repo-root files now match no module and fall through to the existing unmatched handling.

TDD: the first commit adds the failing test against current behavior, the second the fix. Rebased on current main (incl. the path-separator normalization, which touches the same class); `ModuleMapperTest` is 11/11 green on the rebased code.

Pure extension3-internal change (ModuleMapper + its test), no interaction with the currently open PRs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved project change detection so repository-root files, such as `README.md` and `.gitignore`, no longer incorrectly affect the root project.
  * Files within the root project’s source tree continue to be recognized correctly.
  * Nested project matching remains accurate.

* **Tests**
  * Added coverage for repository-root files and root project source changes.

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
@@ -60,7 +60,7 @@ class ModuleMapper {
         for (String changedFile : changedFiles) {
             for (MavenProject project : sortedProjects) {
                 String projectPath = getRelativePath(project, reactorRoot);
-                if (projectPath.isEmpty() ? changedFile.contains("/") : changedFile.startsWith(projectPath + "/")) {
+                if (isFileInProjectSubtree(changedFile, projectPath)) {
                     boolean isTest = isTestPath(changedFile, projectPath);
                     Boolean existing = hasMainChange.get(project);
                     if (existing == null) {
@@ -98,6 +98,15 @@ class ModuleMapper {
             relativeToProject = changedFile.substring(projectPath.length() + 1);
         }
         return relativeToProject.startsWith("src/test/");
+    }
+
+    private static boolean isFileInProjectSubtree(String changedFile, String projectPath) {
+        if (projectPath.isEmpty()) {
+            // Root project: only match files in subdirectories (src/main/...),
+            // not bare repo-root files (README.md, .gitignore)
+            return changedFile.contains("/");
+        }
+        return changedFile.startsWith(projectPath + "/");
     }
 
     private String getRelativePath(MavenProject project, Path reactorRoot) {
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
@@ -60,7 +60,7 @@ class ModuleMapper {
         for (String changedFile : changedFiles) {
             for (MavenProject project : sortedProjects) {
                 String projectPath = getRelativePath(project, reactorRoot);
-                if (projectPath.isEmpty() || changedFile.startsWith(projectPath + "/")) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
## Review Summary

All review questions resolved — no issues found. Clean, well-scoped, well-tested fix.

### What was checked

1. **`changedFile.contains("/")` heuristic** — correctly discriminates bare repo-root files (`README.md`, `.gitignore`) from root-module source files (`src/main/...`). The non-root module branch is unchanged.
2. **No regression for `pom.xml` at root** — verified that `ScalpelLifecycleParticipant` separates POM files into `pomChanges` *before* calling `ModuleMapper`, so the mapper never sees `pom.xml`. Similarly, `.mvn/` files are filtered upstream.
3. **Test coverage** — both the negative case (root files → no match) and the positive case (root source files → still mapped) are covered. TDD approach (failing test first, then fix) is clean.

### Highlights

- Correct root cause identification — the old `projectPath.isEmpty()` catch-all was the bug
- Surgical scope — one extracted `private static` helper, two new tests, zero impact on non-root module matching
- Clear PR description explaining the problem, approach, and test status

---

*This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying.*

famod:
What about files like `.sdkmanrc`, `Jenkinsfile` and `lombok.config`?

gnodet:
Good point — with this change, bare repo-root files like `.sdkmanrc`, `Jenkinsfile`, and `lombok.config` no longer map to the root aggregator, so they won't trigger a rebuild by default.

For `.sdkmanrc` and `Jenkinsfile` that's the right behavior (they're environment/CI config, not module source). But `lombok.config` is a fair call-out since it can affect bytecode generation across all modules.

The escape hatch is `scalpel.fullBuildTriggers` — it runs **before** the module mapper, so matching files trigger a full build without ever reaching this code path. To cover `lombok.config`:

```
-Dscalpel.fullBuildTriggers=.mvn/**,lombok.config
```

or more broadly:

```
-Dscalpel.fullBuildTriggers=.mvn/**,Jenkinsfile,lombok.config
```

(The default is `.mvn/**` — see the [Full Build Triggers](https://github.com/maveniverse/scalpel#full-build-triggers) section in the README.)

---

Worth noting: on older versions (before this fix), `scalpel.excludePaths` can be used as a workaround to stop root files from triggering the root aggregator:

```
-Dscalpel.excludePaths=*.md,LICENSE,.gitignore,.sdkmanrc,.editorconfig
```

The fix here is better because it addresses the root cause — bare root files are structurally not part of any module's subtree — so you don't need to enumerate every new root file (`CODEOWNERS`, `renovate.json`, `SECURITY.md`, …) in the exclude list.

paoloantinori:
Thanks — that's a clear answer. Agreed that `lombok.config` is the interesting one since it affects bytecode generation across modules. Worth a one-line mention in the README's full-build-triggers section as the canonical example of a root-level file teams may want to add; happy to fold it into the docs split (#119) if that lands.

gnodet:
Agreed — a one-liner in the README's full-build-triggers section would help. Suggested wording, right after the `-Dscalpel.fullBuildTriggers=.mvn/**,Jenkinsfile,*.gradle` example:

> **Tip:** Repo-root files (`README.md`, `.gitignore`, `LICENSE`, etc.) are not mapped to any module
> by default. If a root-level file affects your build (e.g. `lombok.config`), add it here so changes
> trigger a full rebuild.

Happy to leave this for the docs split (#119) as @paoloantinori suggested — keeps this PR scoped to the fix.

paoloantinori:
Thanks, that wording is perfect — noted for #119.

famod:
Thanks for the consideration and explanation @gnodet.

So Scalpel is more "optimistic" than GIB in mapping root files, got it!

Issue #50, opened 2026-08-03
External review: functional findings, and an offer to contribute fixes


paoloantinori:
Hi. While adopting Scalpel on Apicurio Registry I ended up doing a systematic pass over 0.3.9, and it turned up more than I could sensibly file one issue at a time. Rather than open a pile of issues and dump triage on you, here is a single summary so you can decide what you actually want.

**I am offering to PR any subset of these.** Tell me which ones you want and in what order, and I will send them one at a time as small, separately reviewable changes with tests. If you would rather I dropped some, that is a completely fine answer too. #49 is already up as the first one.

I verified each item below against the code myself, and I have included how to check it so you do not have to take my word for anything. Where something is a judgement call rather than a defect I have said so.

---

### Correctness

**1. Bare glob patterns only match files at the repository root.**
Path filters compile with `getPathMatcher("glob:" + pattern)` and match the raw relative path, and a single `*` does not cross directory boundaries:

```
glob:*.md    vs README.md                            = true
glob:*.md    vs docs/src/main/asciidoc/foo.md        = false
glob:*.adoc  vs docs/src/main/asciidoc/security.adoc = false
glob:**/*.md vs docs/src/main/asciidoc/foo.md        = true
```

This is live: Quarkus runs Scalpel enabled and its `.mvn/maven.config` sets `-Dscalpel.excludePaths=...,*.md,*.txt,*.adoc`, so nothing under `docs/` is actually excluded and doc-only PRs build more than intended. `README.md` documents the same form (`-Dscalpel.excludePaths=*.md,LICENSE`). Suggested fix is to auto-prefix patterns containing no `/` with `**/`, which makes them mean what people already assume.

**2. `failSafe` does not catch unchecked exceptions.**
`ScalpelCore.detectChanges` catches `IOException | JGitInternalException | ScalpelException`, and `afterProjectsRead` catches `ScalpelException`. Two reachable-from-config paths throw unchecked and hard-fail the build even with `failSafe=true`:

- `Repository.resolve` throws `RevisionSyntaxException` (an `IllegalArgumentException`) for `"  "`, `"@{-99}"`, `"HEAD^{bogus}"`. A stray space in `.mvn/maven.config` is enough, and config values are never trimmed.
- `getPathMatcher` throws `PatternSyntaxException` for `glob:[`, `glob:{a,b`, or a Windows-style `docs\**`.

[...]

PR #67, opened 2026-08-17
fix: validate scalpel config strictly and warn on unknown keys

paoloantinori:
`fromProperties` currently accepts anything: typos in property names (`scalpel.incluedPaths=...`) are silent no-ops, values are never trimmed, and `Boolean.parseBoolean` means `scalpel.enabled=yes` or `=1` silently *disables* the extension.

This adds eager validation in `ScalpelConfiguration`:
- strict boolean parsing (only `true`/`false`, after trimming) with a clear error naming the property;
- value trimming;
- unknown `scalpel.*` key detection with a closest-match suggestion (e.g. `incluedPaths` -> `includePaths`);
- a warnings list surfaced by `ScalpelLifecycleParticipant` right after parsing.

On invalid values `fromProperties` throws `IllegalArgumentException`, which lands in the existing catch in `afterProjectsRead` (warn + build all), so the fail-safe posture you merged for config parsing is unchanged. This was noted as adjacent to #50 item 2 in my follow-up comment there.

Rebase note: one conflict against the config catch block was resolved by keeping the merged catch semantics verbatim and grafting only the warnings loop; the stricter per-value failSafe handling from the original branch was dropped in favour of the current upstream behaviour. Verified: `ScalpelConfigurationTest` 74/74 and `ScalpelLifecycleParticipantTest` 59/59 green on the rebased code.

TDD: first commit is the failing test against current behaviour.

## Summary by CodeRabbit

* **New Features**
  * Configuration values now ignore surrounding whitespace and validate boolean settings consistently.
  * Unknown `scalpel.*` properties are reported with deduplicated warnings and helpful suggestions where applicable.
  * Configuration warnings are displayed during project startup.

* **Bug Fixes**
  * Invalid boolean and safety settings now produce clear, key-specific error messages.

### DIFF
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
@@ -300,7 +300,7 @@ public final class ScalpelConfiguration {
         Set<String> seen = new HashSet<>();
         collectUnknownScalpelKeys(system, seen, warnings);
         collectUnknownScalpelKeys(user, seen, warnings);
-        return warnings.isEmpty() ? Collections.emptyList() : Collections.unmodifiableList(warnings);
+        return warnings.isEmpty() ? List.of() : Collections.unmodifiableList(warnings);
     }
 
     private static void collectUnknownScalpelKeys(Properties props, Set<String> seen, List<String> warnings) {
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
@@ -8,9 +8,12 @@
 package eu.maveniverse.maven.scalpel.core;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Properties;
+import java.util.Set;
 
 public final class ScalpelConfiguration {
 
@@ -57,6 +60,34 @@ public final class ScalpelConfiguration {
     private static final String DEFAULT_REPORT_FILE = "target/scalpel-report.json";
     public static final long DEFAULT_MAX_RESOURCE_FILE_SIZE = 10L * 1024 * 1024; // 10 MB
 
+    private static final Set<String> KNOWN_KEYS = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(
+            ENABLED,
+            BASE_BRANCH,
+            HEAD,
+            ALSO_MAKE,
+            ALSO_MAKE_DEPENDENTS,
+            FULL_BUILD_TRIGGERS,
+            FAIL_SAFE,
+            MODE,
+            DISABLE_ON_BRANCH,
+            DISABLE_ON_BASE_BRANCH,
+            EXCLUDE_PATHS,
+            INCLUDE_PATHS,
+            DISABLE_TRIGGERS,
+            DISABLE_ON_SELECTED_PROJECTS,
+            SKIP_TESTS_FOR_UPSTREAM,
+            SKIP_TESTS_FOR_DOWNSTREAM_MODULES,
+            UPSTREAM_ARGS,
+            DOWNSTREAM_ARGS,
+            FETCH_BASE_BRANCH,
+            UNCOMMITTED,
+            UNTRACKED,
+            FORCE_BUILD_MODULES,
+            BUILD_ALL_IF_NO_CHANGES,
+            IMPACTED_LOG,
+            REPORT_FILE,
+            MAX_RESOURCE_FILE_SIZE)));
+
     private final boolean enabled;
     private final String baseBranch;
     private final String head;
@@ -83,6 +114,7 @@ public final class ScalpelConfiguration {
     private final String mode;
     private final String reportFile;
     private final long maxResourceFileSize;
+    private final List<String> warnings;
 
     private ScalpelConfiguration(
             boolean enabled,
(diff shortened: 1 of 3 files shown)


cstamas:
Changed my mind: targeting 3.10+ only, will leave 3.9 for tricks with split. I just realized something... Once merged. 3.9 support module is to be dropped.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
```diff
@@ -243,6 +284,80 @@ private static List<String> parseList(String value) {
...
+        for (String known : KNOWN_KEYS) {
+            int dist = editDistance(unknown, known);
+            if (dist < bestDist) {
+                bestDist = dist;
+                best = known;
+            }
+        }
+        // Only suggest if the edit distance is at most half the suffix length
+        // (after stripping the shared "scalpel." prefix). Without this adjustment,
+        // the 8-char shared prefix inflates the threshold and causes unrelated keys
```
gnodet, line 331: **Nit (non-blocking):** `best.length()` is evaluated here before the `best != null` guard on the next line. Safe in practice since `KNOWN_KEYS` is never empty, but flipping the order would be marginally more defensive:

```java
if (best == null) {
    return null;
}
int suffixLen = Math.max(unknown.length(), best.length()) - PREFIX.length();
if (bestDist <= suffixLen / 2) {
    return best;
}
return null;
```

gnodet (approved):
## Review Summary

Clean, well-scoped PR that fixes a real usability problem — `Boolean.parseBoolean` silently treating `scalpel.enabled=yes` or `=1` as `false` was a nasty gotcha. The strict validation, value trimming, and unknown-key detection with edit-distance suggestions are all solid additions.

### What was checked

**Q1: Does strict boolean parsing preserve fail-safe behavior?**
✅ Yes — the `afterProjectsRead` catch block (line 88) catches `Exception` broadly, so `IllegalArgumentException` from `parseStrictBoolean` is caught, logged as a warning, and the build proceeds normally. The fail-safe posture is unchanged.

**Q2: Is the `KNOWN_KEYS` set complete?**
✅ Yes — 26 `scalpel.*` constants match 26 entries in `KNOWN_KEYS` exactly 1:1. No key is missing or extra.

**Q3: Is the edit distance + suggestion threshold correct?**
✅ Yes — standard Levenshtein implementation, and the suffix-aware threshold (`bestDist <= suffixLen / 2`) correctly suggests for near-typos (`scalpel.enbaled` → `scalpel.enabled`) while suppressing misleading suggestions for unrelated keys (`scalpel.foobar`).

### Minor suggestions (non-blocking)

1. **Test gap:** `unknownKey_completelyWrong_producesWarning` asserts the warning exists but doesn't assert that no "Did you mean" suggestion is included — adding that would explicitly verify the threshold suppression.

2. **Defensive ordering in `findClosestKey`:** `best.length()` is evaluated before the `best != null` guard. Unreachable in practice (KNOWN_KEYS has 26 entries), but moving the null check first would be cleaner.

Excellent test coverage (131 new lines of parameterized tests) and clean TDD approach. Nice work! 👏

---

core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelConfigurationTest.java
```diff
@@ -452,4 +453,135 @@ void listField_trimsWhitespaceInValues() {
...
+    }
+
+    @Test
+    void unknownKey_completelyWrong_producesWarning() {
+        Properties sys = new Properties();
+        sys.setProperty("scalpel.foobar", "x");
+        ScalpelConfiguration config = ScalpelConfiguration.fromProperties(sys, new Properties());
+        List<String> warnings = config.getWarnings();
+        assertFalse(warnings.isEmpty());
+        assertTrue(warnings.stream().anyMatch(w -> w.contains("scalpel.foobar")));
```
gnodet, line 555: **Suggestion (non-blocking):** This test verifies the warning is produced for a completely wrong key, but doesn't assert that no "Did you mean '...'" suggestion is included. Adding a negative assertion would explicitly verify the threshold suppression logic:

```suggestion
        assertTrue(warnings.stream().anyMatch(w -> w.contains("scalpel.foobar")));
        assertTrue(warnings.stream().noneMatch(w -> w.contains("Did you mean")),
                "completely wrong key should not produce a suggestion");
```

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
paoloantinori: Applied in 83fd926: early `return null` guard before evaluating `best.length()`.

core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelConfigurationTest.java
paoloantinori: Applied in 83fd926: added `noneMatch(w -> w.contains("Did you mean"))` to `unknownKey_completelyWrong_producesWarning`.

paoloantinori:
Thanks @cstamas — good to know about the 3.10+ retarget. For the record, this PR only touches configuration parsing with plain `java.util` (no Maven 3.9-specific APIs), so dropping the 3.9 support module won't affect it. Thanks @gnodet for the review — both non-blocking suggestions applied in 83fd926.

cstamas:
@paoloantinori sorry, my comment was a copy and pasta into **wrong issue** (and even project) :smile: So, not for scalpel, this is why I hide it as "out of scope/off-topic". Sorry again.

paoloantinori:
@cstamas no worries — thanks for the heads-up. Ignoring the 3.10+ note then; nothing to change on our side either way.

PR #128, opened 2026-08-24
Fix #81: user properties take precedence over system properties, per Maven convention

### LINKED ISSUE #81: [bug] Configuration precedence inverts the Maven convention: MAVEN_OPTS silently overrides explicit -D arguments
## Behavior

`ScalpelConfiguration.resolve()` checks system properties first and user properties second, but the caller passes `session.getSystemProperties()` as system and `session.getUserProperties()` as user.

In Maven, CLI `-D` arguments land in **user** properties, while JVM properties (e.g. anything in `MAVEN_OPTS`) land in **system** properties. So today:

```bash
MAVEN_OPTS=-Dscalpel.mode=report mvn -Dscalpel.mode=trim verify
```

silently runs in `report` mode. An explicit command-line argument loses to an environment-level setting, which is the opposite of standard Maven precedence. Entries from `.mvn/maven.config` become CLI args and therefore also lose to `MAVEN_OPTS`.

## Expected

User properties take precedence over system properties, matching the rest of Maven.

Separately, project POM `<properties>` are not consulted at all. That looks deliberate (a PR author cannot reconfigure Scalpel through `pom.xml`) and I think it should stay that way, but it would be worth documenting the precedence order and that boundary explicitly in the README.

## Suggested fix

Swap the lookup order in `resolve()`, add a test where both maps carry different values, and document the precedence. Happy to PR this with a failing test first if the change is welcome.

paoloantinori:
## Problem

`ScalpelConfiguration.resolve()` checked system properties before user properties. Since the caller passes `session.getSystemProperties()` / `session.getUserProperties()`, and in Maven CLI `-D` lands in **user** properties while JVM properties (e.g. `MAVEN_OPTS`) land in **system** properties, an environment-level setting silently overrode an explicit command-line argument:

```bash
MAVEN_OPTS=-Dscalpel.mode=report mvn -Dscalpel.mode=trim verify   # ran in report mode
```

`.mvn/maven.config` entries become CLI args and also lost to `MAVEN_OPTS`.

## Change

- `resolve()` now checks user first, then system, then default. All 25 properties go through this one helper; there is exactly one production caller of `fromProperties`, and CI base-branch detection reads `env.*` keys directly from the system map, so nothing else changes.
- README documents the precedence order and that POM `<properties>` are deliberately not read (a PR cannot reconfigure Scalpel through `pom.xml`).

Note: anyone relying on `MAVEN_OPTS=-Dscalpel.enabled=false` as a global kill switch overriding an explicit CLI value will see the CLI win after this change; that inversion is the point of #81.

## Verification

RED commit first (test failing on unmodified main), then the fix. Full `core`+`extension3` suites green (146 + 75 tests). Includes a test pinning `-Dkey=` empty-string semantics (present-but-empty wins, matching Maven).

## Summary by CodeRabbit

* **Documentation**
  * Clarified configuration precedence: user properties override system properties, which override defaults.
  * Documented that project POM properties do not configure Scalpel.

* **Bug Fixes**
  * Updated configuration handling so user-supplied values take precedence, including explicitly empty values.


gnodet (approved):
Clean, well-tested bug fix that corrects an inverted property precedence in `ScalpelConfiguration.resolve()`. The change aligns Scalpel with Maven's standard convention where user properties (CLI `-D`) take precedence over system properties (JVM/MAVEN_OPTS). The code change is minimal and surgical, the tests cover both the main case and an important edge case (empty-string semantics), and the README documentation is clear and accurate.

### General Comments

- The red-green testing approach (failing test first, then the fix) is excellent engineering practice.
- The `emptyUserPropertyWinsOverSystemProperty` test is a particularly thoughtful addition — empty-but-present vs. absent is a subtle distinction in `Properties`-based config, and pinning it here prevents future regressions.
- The README documentation concisely explains the precedence chain and the deliberate exclusion of POM `<properties>`.
- The PR description is exemplary: clear problem statement, concise change summary, explicit acknowledgment of the behavioral change for existing users.

### 📋 PR Metadata

---

PR #133, opened 2026-08-25
Fix #126: unordered list comparison must not collapse duplicate entries

### LINKED ISSUE #126: [arch] PomChangeAnalyzer mixes diffing, topology and filesystem I/O; accumulator out-parameters create ordering hazards
## Problem

PomChangeAnalyzer is ~1200 lines mixing pure model diffing, reactor topology and filesystem I/O. The parameter-passing style is the harder problem: `analyzeParentPomChange` takes nine parameters, four of them mutable accumulator sets mutated in place across a 224-line method, so correctness depends on statement ordering (one augmentation must run before an `addAll`, and the child loop mutates the managed-dependency accumulator again). `analyzeProfileChanges` has the same shape.

One low-severity correctness bug in here: `equalStringListsUnordered` compares two hash sets after a size check, returning true for `[a, a, b]` vs `[a, b, b]` (affects resource includes/excludes with duplicates). Comparing sorted copies fixes it.

Separating pure diffing from I/O is also what makes the resource-scanning and POM-text memoization optimizations tractable, and enables property-based testing of the differ.

## Proposal

Extract ReactorTopology, ModelDiffer (returning a `ModelDiff` value object instead of accumulator out-params), Xpp3Doms utility, PropertyReferences, and FilteredResourceScanner (the only I/O part). Fix the duplicate-collapsing comparison. Behavior unchanged per the existing test cases.

paoloantinori:
## Problem

As filed in #126 (bug subset): `PomChangeAnalyzer.equalStringListsUnordered` compared two `HashSet`s after a size check, so `[xml, xml, properties]` vs `[xml, properties, properties]` compared equal. A change to a filtered resource's `includes`/`excludes` that only alters DUPLICATED entries was therefore invisible: the resource looked unchanged and the module was not marked affected (under-build).

## Change

Compare sorted copies instead of set views: unordered semantics preserved, duplicates respected. The method's contract and callers (resource includes/excludes comparison) are unchanged.

## Verification

TDD: RED first (two new parameterized cases, duplicated-include and duplicated-exclude, failing exactly as predicted), then the fix. Full `core` + `extension3` suites green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved comparison of unordered resource include and exclude entries.
  * Reordered entries with duplicates are now correctly treated as equivalent, preventing unnecessary parent POM changes.

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
@@ -18,7 +18,6 @@ import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -809,7 +808,11 @@ class PomChangeAnalyzer {
         if (a.size() != b.size()) {
             return false;
         }
-        return new HashSet<>(a).equals(new HashSet<>(b));
+        List<String> sortedA = new ArrayList<>(a);
+        List<String> sortedB = new ArrayList<>(b);
+        sortedA.sort(String::compareTo);
+        sortedB.sort(String::compareTo);
+        return sortedA.equals(sortedB);
     }
 
     private boolean equalRepositoryLists(List<Repository> a, List<Repository> b) {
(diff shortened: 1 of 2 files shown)


gnodet (approved):
Clean, well-tested bugfix for the duplicate-collapsing comparison in `equalStringListsUnordered`. The sorted-list approach correctly preserves unordered semantics while respecting duplicates, and the TDD commit structure (RED-first test proving the bug, then the fix) is solid.

Strengths worth calling out:
- Both callers of `equalStringListsUnordered` (resource includes and excludes) have dedicated parameterized test cases covering the exact duplicate-redistribution pattern (`[a, a, b]` vs `[a, b, b]`)
- The `HashSet` import removal is clean — no other usage remains in the file
- CI green on all 9 matrix jobs

### 📋 PR Metadata

PR #134, opened 2026-08-25
Fix #102, fix #103: pin actions to SHAs, permissions blocks, quoted version expansion

### LINKED ISSUE #102: [security] Pin GitHub Actions to commit SHAs: release job passes Maven Central identity to a floating third-party tag
## Problem

`.github/workflows/release.yml` passes the project's publishing identity (`GPG_SIGNING_KEY`, `GPG_PASSPHRASE`, `MAVEN_USER`/`MAVEN_PASSWORD`) to `crazy-max/ghaction-import-gpg@v7`, a mutable tag in a third-party repository. If that repository or the maintainer's account is compromised, the repointed tag executes in the release job with those secrets in the environment, allowing arbitrary artifacts to be signed and published to Maven Central under the `eu.maveniverse` identity. Since Scalpel is consumed by Quarkus, Camel and others, this compromises downstream consumers, not just this CI.

The same applies to `actions/checkout@v6` / `actions/setup-java@v5` in all workflows and to the reusable `maveniverse/parent` workflow reference in `ci.yml`.

## Proposal

- Every `uses:` references a full 40-character commit SHA with a version comment; the GPG import action first as the highest-risk one.
- Dependabot's `github-actions` ecosystem enabled so the SHAs stay current.
- A one-line contributing note stating the pinning requirement for new workflow steps.

Happy to PR this.

### LINKED ISSUE #103: [security] Missing permissions blocks in ci.yml/release.yml and unquoted template expansion in release.yml
## Problem

1. `ci.yml` and `release.yml` declare no `permissions:` block, so `GITHUB_TOKEN` gets the repository default, which on older repositories is read/write across all scopes; `ci.yml` also delegates to a reusable workflow in the parent repo which inherits those permissions. `deploy-snapshot.yml` and `sonarcloud.yml` already set `contents: read` correctly, so this is filling two gaps, not a new practice.
2. `release.yml` assigns an unquoted template expansion (`version=${{ steps.version.outputs.version }}`) into a shell script. The tag trigger glob matches far more than semver, so a tag containing a command substitution flows into the shell of the job holding the GPG key. Requires write access (not a fork vector), but it is a cheap hardening; `sonarcloud.yml` already demonstrates the correct `env:` + quoted-variable pattern.

## Proposal

- `permissions: contents: read` in ci.yml; minimal exact permissions in release.yml.
- Pass the tag version through `env:` and reference it as a quoted shell variable.
- Tighten the tag trigger glob.

Happy to PR this together with the action-pinning issue or separately.

paoloantinori:
## #102 Pin GitHub Actions to commit SHAs

Every `uses:` now references a full 40-char commit SHA with a version comment (all verified against the tags via the GitHub API):

| Action | Pinned | Tag |
|---|---|---|
| actions/checkout | `3d3c42e5` | v7 |
| actions/setup-java | `b6effb05` | v5.7.0 |
| crazy-max/ghaction-import-gpg | `2dc316de` | v7.0.0 |

The GPG import action runs in the release job holding the Maven Central signing and publishing secrets; a mutable tag there is the highest-impact unpinned reference in the repo. The reusable `maveniverse/parent` workflow ref is left on its branch-pin convention (noted, not changed).

## #103 Permissions and quoting

- `ci.yml`: explicit `permissions: contents: read` (had none; it inherits the parent workflow's usage).
- `release.yml`: explicit minimal `permissions: contents: write, issues: write` — `gh release create` needs contents:write, the milestone commands need the issues scope; nothing else in the job uses GITHUB_TOKEN.
- `release.yml` Post-release step: the tag version is now passed via `env:` (`VERSION`) and assigned as a quoted shell variable, matching the pattern already used in sonarcloud.yml, instead of an unquoted `${{ }}` expansion into the shell.

All four workflow files parse as valid YAML; no floating `uses:` remains except the intentional parent-workflow branch ref.

### DIFF
--- a/.github/workflows/deploy-snapshot.yml
+++ b/.github/workflows/deploy-snapshot.yml
@@ -15,12 +15,12 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           fetch-depth: 0
 
       - name: Setup Java
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5.7.0
         with:
           distribution: 'temurin'
           java-version: '21'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
       - main
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     name: Verify
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
     steps:
 
       - name: Checkout ${{ github.ref_name }}
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
         with:
           fetch-depth: 0
           ref: ${{ github.ref }}
@@ -41,14 +41,14 @@ jobs:
           echo "version=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
 
       - name: Setup Java
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5.7.0
         with:
           distribution: temurin
           java-version: 21
           cache: maven
 
       - name: Import GPG key
-        uses: crazy-max/ghaction-import-gpg@v7
+        uses: crazy-max/ghaction-import-gpg@2dc316deee8e90f13e1a351ab510b4d5bc0c82cd # v7.0.0
         with:
           gpg_private_key: ${{ secrets.GPG_SIGNING_KEY }}
           passphrase: ${{ secrets.GPG_PASSPHRASE }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,10 @@ on:
     tags:
       - '[0-9]*.[0-9]*.[0-9]*'
 
+permissions:
+  contents: write
+  issues: write
+
 jobs:
   release:
     name: Release to Maven Central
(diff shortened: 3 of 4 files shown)


.github/workflows/deploy-snapshot.yml
```diff
@@ -15,12 +15,12 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
```
gnodet, line 18: **Low:** The SHA resolves to `v7.0.1`, but the comment says `# v7`. Other pins in this PR use the specific point version (`# v5.7.0`, `# v7.0.0`). Consider updating for consistency:

```suggestion
        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```

gnodet (approved):
Well-structured security hardening PR that pins all direct GitHub Actions to verified commit SHAs, adds minimal permissions blocks, and fixes an unquoted template expansion in the release workflow. All three SHA pins verified against the GitHub API.

**Minor observation (non-blocking):**

The version comment on `actions/checkout` says `# v7` (major-only), while all other pins use specific point versions (`# v5.7.0`, `# v7.0.0`). The SHA `3d3c42e5...` actually resolves to `v7.0.1`. Using the specific version makes it easier for humans and Dependabot to track what is pinned.

### 📋 PR Metadata

| Category | _(unlabeled)_ | `chore` |

gnodet (commented):
## Re-review (post-merge)

This PR was merged while the re-review was in progress. Flagging a regression introduced by the merge conflict resolution for a follow-up fix.

### Verified Findings

| # | Severity | File | Finding | Verdict |
|---|----------|------|---------|---------|
| 1 | 🔴 high | `release.yml` | Merge conflict resolution reverted release-drafter integration from PR #146 | ✅ confirmed |
| 2 | 🟡 low | `deploy-snapshot.yml` | `actions/checkout` comment says `# v7` instead of `# v7.0.1` (inconsistent with other pins) | ✅ confirmed |

### 🔴 Release-drafter regression

The merge-with-main commit resolved a conflict in `release.yml` by dropping the conditional release-drafter logic that PR #146 added:

**PR #146 added** (now on `main` but reverted by this merge):
```yaml
if gh release view "$version" > /dev/null 2>&1; then
  gh release edit "$version" --draft=false --verify-tag --title "$name $version"
else
  gh release create "$version" --verify-tag --notes-from-tag --title "$name $version"
fi
```

**This PR restored** (pre-#146 pattern):
```yaml
gh release create "$version" --verify-tag --notes-from-tag --title "$name $version" || echo "skipping"
```

Since the release-drafter workflow and config from PR #146 still exist on `main`, draft releases will be created but never published — `gh release create` will fail when a draft already exists for the tag, and `|| echo` silently swallows the error.

**Suggested follow-up:** Restore the conditional from PR #146 in a new commit on `main`.

PR #135, opened 2026-08-25
Fix #85: validate fetch remote against configured remotes, handle slash-containing branch names

### LINKED ISSUE #85: [bug] fetchBranch treats an unknown remote name as a URL: crafted baseBranch fetches from an attacker-chosen host; also mis-splits release/1.0
## Behavior

`GitChangeDetector.fetchBranch` splits `baseBranch` at the first slash and passes the left half straight to `git.fetch().setRemote(remote)`. JGit treats an unknown remote *name* as a URL (verified against the jgit 5.13.5 jar in use: `Transport.open` falls back to constructing a `URIish` from the string when no remote config exists).

Two consequences:

1. **Crafted config fetches from an arbitrary host.** A `scalpel.baseBranch` of `git@attacker.example:evil.git/main` splits into remote `git@attacker.example:evil.git` and branch `main`; since no configured remote has that name, JGit parses it as an scp-style URI and fetches from the attacker's host using the runner's SSH agent or credential helper. Beyond credential exposure, the attacker then controls the objects in the base ref used for diffing, so they can forge the changed-file set and hide modules from the build. A wildcard in the branch position silently widens the refspec to fetch every branch. The CI auto-detect path is safe (`detectBaseBranch` hardcodes the `origin/` prefix); the risk is explicit configuration. This is a Scalpel logic issue, not a JGit vulnerability.

2. **Plain bug in the same method:** splitting at the *first* slash means a legitimate local base branch named `release/1.0` is parsed as remote `release`, branch `1.0`, and the fetch fails.

## Expected

- The remote is rejected unless `repository.getRemoteNames()` contains it.
- The branch is validated with `Repository.isValidRefName`; wildcards rejected.
- A base branch with no remote prefix is handled correctly rather than mis-split at the first slash.
- A fetch timeout so an unreachable remote cannot hang the build indefinitely.

## Suggested fix

I can PR this with failing tests first covering a URL-shaped remote, a wildcard branch, and a slash-containing local branch name. Happy to proceed if welcome.

paoloantinori:
## Problem

As filed in #85: `fetchBranch` split `baseBranch` at the first slash and passed the left half to `git.fetch().setRemote(...)`. JGit still treats an unknown remote name as a URL (verified on JGit 7.7.1: the RED test shows `git@host:evil.git/main` entering the fetch path as a URI), so a crafted `scalpel.baseBranch` could fetch from an attacker-chosen host using the runner's credentials, and a wildcard branch silently widened the refspec. Separately, a legitimate local branch named `release/1.0` was mis-split into remote `release` / branch `1.0` and the fetch failed.

## Change

- URL-shaped values (anything containing `:`) are rejected up front with an IOException naming the configuration problem, before any transport opens. Git refnames cannot contain `:` (git-check-ref-format), so no legitimate value is rejected.
- The pre-slash prefix is treated as a remote ONLY if `repository.getRemoteNames()` contains it; otherwise the whole value is a local branch name and the fetch is skipped with a WARN (visibility matters here: CI auto-detection hardcodes `origin/`, and this path is only reached when the ref already fails to resolve — previously an unconfigured origin failed loudly, so the WARN keeps the situation visible instead of a silent debug-level skip).
- The branch is validated with `Repository.isValidRefName("refs/heads/" + branch)`, which also rejects wildcards.

Deliberate behavior changes (two #63-era tests updated accordingly): `bogus/main` with no remote named `bogus` is now treated as a local branch (skip + WARN) instead of throwing from a bogus fetch; `:/nope` is rejected as URL-shaped instead of escaping as a raw IllegalArgumentException.

## Verification

TDD: RED first (URL-shaped, mis-split and wildcard tests failing on main), then the fix, then the review follow-up (WARN test + guard simplification). Full `core` (135) + `extension3` (146) suites green, offline (`-o`) verified: no test touches the network.

### DIFF
--- a/core/src/test/java/eu/maveniverse/maven/scalpel/core/GitChangeDetectorTest.java
+++ b/core/src/test/java/eu/maveniverse/maven/scalpel/core/GitChangeDetectorTest.java
@@ -15,7 +15,9 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -455,6 +457,43 @@ class GitChangeDetectorTest {
         try (Repository repo = repoWithOriginRemote()) {
             // "release" is not a configured remote: the whole string is a local branch name
             assertDoesNotThrow(() -> detector.fetchBranch(repo, "release/1.0"));
+        }
+    }
+
+    /**
+     * Runs the callable with System.err captured (slf4j-simple writes to stderr)
+     * and returns everything that was written during its execution.
+     */
+    private String captureStderr(Runnable action) {
+        PrintStream originalErr = System.err;
+        ByteArrayOutputStream captured = new ByteArrayOutputStream();
+        System.setErr(new PrintStream(captured, true, StandardCharsets.UTF_8));
+        try {
+            action.run();
+        } finally {
+            System.setErr(originalErr);
+        }
+        return captured.toString(StandardCharsets.UTF_8);
+    }
+
+    @Test
+    void fetchBranch_unconfiguredPrefixWithSlash_warns() throws Exception {
+        try (Repository repo = repoWithOriginRemote()) {
+            String err = captureStderr(() -> assertDoesNotThrow(() -> detector.fetchBranch(repo, "bogus/main")));
+            assertTrue(
+                    err.contains("WARN "),
+                    "expected a WARN about the unconfigured remote prefix but stderr was: " + err);
+            assertTrue(
+                    err.contains("'bogus' is not a configured remote"),
+                    "WARN should name the unconfigured prefix 'bogus' but stderr was: " + err);
+        }
+    }
+
+    @Test
+    void fetchBranch_plainBranchNoSlash_noWarn() throws Exception {
+        try (Repository repo = repoWithOriginRemote()) {
+            String err = captureStderr(() -> assertDoesNotThrow(() -> detector.fetchBranch(repo, "main")));
+            assertFalse(err.contains("WARN "), "no WARN expected for a plain branch name but stderr was: " + err);
         }
     }
 
--- a/core/src/test/java/eu/maveniverse/maven/scalpel/core/GitChangeDetectorTest.java
+++ b/core/src/test/java/eu/maveniverse/maven/scalpel/core/GitChangeDetectorTest.java
@@ -410,8 +410,9 @@ class GitChangeDetectorTest {
     // ---- fetchBranch validation (#85) ----
 
     /**
-     * Creates a work repo with a commit on branch main, plus a bare remote named "origin"
-     * that has main pushed to it, and configures the remote on the work repo.
+     * Creates a work repo with a commit on its initial branch (whatever the environment
+     * defaults to), plus a bare remote named "origin" that has that branch pushed to it,
+     * and configures the remote on the work repo.
      */
     private Repository repoWithOriginRemote() throws Exception {
         Path remoteDir = tempDir.resolve("remote.git");
@@ -427,8 +428,11 @@ class GitChangeDetectorTest {
                 Files.write(workDir.resolve("file.txt"), "hello".getBytes(StandardCharsets.UTF_8));
                 git.add().addFilepattern("file.txt").call();
                 git.commit().setMessage("initial").call();
-                git.branchCreate().setName("main").call();
-                git.push().setRemote(remoteDir.toUri().toString()).add("main").call();
+                // The initial branch name comes from the environment (init.defaultBranch),
+                // so derive it from the repository instead of creating "main", which
+                // already exists where that is the default
+                String branch = git.getRepository().getBranch();
+                git.push().setRemote(remoteDir.toUri().toString()).add(branch).call();
                 git.remoteAdd()
                         .setName("origin")
                         .setUri(new org.eclipse.jgit.transport.URIish(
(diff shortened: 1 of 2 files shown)


core/src/test/java/eu/maveniverse/maven/scalpel/core/GitChangeDetectorTest.java
```diff
@@ -403,4 +406,119 @@ void readPomFilesAtCommit_readsMultiplePoms() throws Exception {
...
+        try (Git remoteGit =
+                Git.init().setDirectory(remoteDir.toFile()).setBare(true).call()) {
+            try (Git git = Git.init().setDirectory(workDir.toFile()).call()) {
+                git.getRepository().getConfig().setString("user", null, "name", "Scalpel Test");
+                git.getRepository().getConfig().setString("user", null, "email", "scalpel@test.invalid");
+
+                Files.write(workDir.resolve("file.txt"), "hello".getBytes(StandardCharsets.UTF_8));
+                git.add().addFilepattern("file.txt").call();
+                git.commit().setMessage("initial").call();
+                git.branchCreate().setName("main").call();
```
gnodet: **Medium (non-blocking):** This hardcodes `"main"` which will throw `RefAlreadyExistsException` if the test environment has `init.defaultBranch=main`. The existing `fetchBranch_successfulFetchFromBareRemote` test at line 323 demonstrates the robust pattern:

```suggestion
                String initialBranch = git.getRepository().getBranch();
                git.push().setRemote(remoteDir.toUri().toString()).add(initialBranch).call();
```

gnodet (approved):
Well-crafted security hardening that correctly closes the URL-injection vector in `fetchBranch` with layered validation (URL check, remote allowlist, refname validation), properly handles slash-containing branch names, and ships with thorough TDD-style test coverage.

The validation layering is well-designed: the colon check is the cheapest gate and catches all URL schemes; the remote allowlist via `getRemoteNames()` is the core security control; `isValidRefName` is the safety net catching wildcards and malformed refnames.

**Observations (non-blocking):**

1. **Test helper `repoWithOriginRemote()`** — hardcodes `git.branchCreate().setName("main")`, which may throw `RefAlreadyExistsException` if the test environment has `init.defaultBranch=main`. The existing `fetchBranch_successfulFetchFromBareRemote` test demonstrates the robust pattern: use `git.getRepository().getBranch()` to discover the actual initial branch.

2. **Use-after-close pattern** — `repoWithOriginRemote()` returns `git.getRepository()` inside a try-with-resources that closes the Git object. Technically use-after-close, though it works in practice for config reads.

3. **Stderr capture coupling** — The `captureStderr` approach for WARN log assertions couples tests to `slf4j-simple` writing to `System.err`. Pragmatic but fragile.

### 📋 PR Metadata

gnodet (approved):
## Review Summary

Well-designed security fix that properly closes the URL-injection vector in `fetchBranch` with layered validation (colon/URL rejection, remote allowlist via `getRemoteNames()`, refname validation via `isValidRefName`). The fix also correctly handles slash-containing local branch names like `release/1.0`. Test coverage is thorough with 8 new/updated tests, CI green across all platform/JDK combinations.

### Observations (non-blocking)

- The validation layering is well-ordered: colon check is O(1) and catches all URL schemes; remote allowlist is the core security control; `isValidRefName` is the safety net rejecting wildcards and malformed names.
- The behavioral change from throwing on `bogus/main` to treating it as a local branch (skip + WARN) is the right call for #85.
- Minor: `repoWithOriginRemote()` hardcodes `"main"` for `branchCreate()` which could be fragile if JGit defaults change — the existing `fetchBranch_successfulFetchFromBareRemote` test shows the safer pattern using `git.getRepository().getBranch()`.

### 📋 PR Metadata

| Labels | _(none)_ | + `bug` |

---

core/src/test/java/eu/maveniverse/maven/scalpel/core/GitChangeDetectorTest.java
paoloantinori: Fixed in 1f74edd: `repoWithOriginRemote()` now derives the branch from the repository after the initial commit (`git.getRepository().getBranch()`) and pushes that, instead of creating and pushing `"main"`, the same pattern as `fetchBranch_successfulFetchFromBareRemote`.

Verified the failure mode by running the suite with a simulated `init.defaultBranch=main` global config (JGit picks it up from the environment): all six tests using the helper errored with `RefAlreadyExistsException` before the change, and all pass after. The other new tests do not depend on the fixture branch name (`fetchBranch` returns early for slash-free names before any ref lookup), so no other spot in this PR needed the same treatment.

gnodet (approved):
## Re-review (1 new commit: derive fixture branch name)

The commit correctly addresses the previous review feedback by deriving the fixture's branch name from the repository via `git.getRepository().getBranch()` instead of hardcoding `"main"`, making the test robust across environments with different `init.defaultBranch` settings.

The fix removes the now-unnecessary `branchCreate()` call and updates the Javadoc accordingly. All six tests using `repoWithOriginRemote()` were verified — none depend on the fixture's branch being named `"main"`.

+8/-4 lines, purely in the test fixture — minimal scope, directly addressing the feedback. 👍

### 📋 PR Metadata

---

gnodet (approved):
Well-designed security fix that correctly prevents crafted `scalpel.baseBranch` from fetching arbitrary remote URLs via JGit's fallback behavior. The defense-in-depth approach (colon rejection, remote name validation, branch name validation via `isValidRefName`) is sound, no legitimate configuration is rejected, and test coverage is comprehensive with TDD methodology.

**Observations (non-blocking):**
- The `repoWithOriginRemote()` test helper creates a `Git` object via `Git.init()` inside try-with-resources, then returns `git.getRepository()`. When the try block exits, `Git.close()` sets the repository's use-count to zero. Tests pass because JGit's `FileRepository` is resilient for the operations used, but the lifecycle is technically incorrect.
- The `captureStderr()` helper is not thread-safe (redirects `System.err`). Works because Surefire runs sequentially, but a comment noting this assumption would help future maintainers.
- No test for the new null/empty `baseBranch` early-return guard. The caller (`ScalpelCore`) already null-checks, so this is a defensive belt-and-suspenders measure.
- The TDD commit structure (RED tests first, then fix) is excellent.
- Note: the `disableOnBaseBranch` logic in `ScalpelCore.java` still uses naive first-slash splitting for pattern matching — pre-existing and outside this PR's scope.

### 📋 PR Metadata

PR #137, opened 2026-08-25
Fix #89: report writes honour failSafe; every bail-out overwrites stale reports with a status document

### LINKED ISSUE #89: [bug] Report writes ignore failSafe and stale reports survive failed runs
## Behavior

Two related problems in the failure path of report mode, the mode that is supposed to observe without disturbing the build.

**Writes ignore failSafe.** The report and impacted-log write sites throw `MavenExecutionException` unconditionally. A read-only `target/` or an unwritable `reportFile` path fails the build even with `failSafe=true`.

**Stale reports survive.** When `detectChanges` returns null under failSafe, the code returns before `writeReport`, so nothing invalidates a previous `target/scalpel-report.json`. On any reused workspace, CI tooling reads stale data believing it is current. The JSON carries no timestamp, no head SHA and no status field, so a consumer cannot even detect the staleness.

Related ordering hazard: in trim mode `session.setProjects(buildSet)` happens before `applyPerCategoryArgs`, so an exception between the two leaves a trimmed reactor with no args applied.

## Expected

- Write failures honour `failSafe` instead of always throwing.
- Every failSafe bail-out overwrites the previous report with an explicit status document.
- The report carries `headSha`, `mergeBase` and a generation timestamp so consumers can detect staleness.
- Trim-mode reactor mutation and per-category args applied atomically (compute everything, then apply).

Happy to PR with failing tests first if the direction is welcome.

paoloantinori:
## Problem

As filed in #89, two hazards in the report-mode failure path:

1. **Writes ignored failSafe.** The report and impacted-log write sites threw `MavenExecutionException` unconditionally, so a read-only `target/` failed the build even with `scalpel.failSafe=true`.
2. **Stale reports survived.** Every bail-out path returned before writing, so on a reused workspace CI kept reading the previous run's `target/scalpel-report.json` as if current - including the most common path of all: a run with no changes.

## Change

- `handleWriteFailure`: with failSafe=true a write failure WARNs and the build continues; with failSafe=false it throws as before. Covers the JSON report, the full-build report and the impacted log.
- Every report-mode bail-out now overwrites the configured reportFile with a minimal status document instead of leaving the previous run's file: new optional `status` (failed/skipped) and `reason` fields. `ScalpelCore` records why detection did not run, so deliberate conditions (`disableOnBranch`/`disableOnBaseBranch` match, not a git repo, no base branch) are reported as `status:"skipped"` with the exact reason, while genuine failures (failSafe bail-outs, unexpected exceptions) are `status:"failed"`. No-changes, all-excluded and disableTriggers paths get accurate skipped reasons too.
- Schema (`scalpel-report-v2.schema.json`) and README document the optional status/reason fields; additive only, no version bump per the #60 policy.

**Deliberate behavior changes** (flagging explicitly): report mode now ALWAYS produces a report file at the configured path (status document at minimum) - code relying on "no report file = nothing detected" must read `status` instead; one existing test was updated accordingly.

## Verification

TDD across three commits (RED failSafe/stale-report tests, GREEN implementation, review follow-up covering the disable-reason routing and the remaining bail-outs). Full suites green (core 129, extension3 152).

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,14 @@ is published alongside the code. The current version is **2**.
 }
 ```
 
+**Status-only reports:** when analysis does not complete (failSafe bail-out, unexpected
+error) or is deliberately skipped (no changes detected, disabled by
+`disableOnBranch`/`disableOnBaseBranch`/`disableTriggers`, or all changed files excluded
+by path filters), Scalpel overwrites the report file with a minimal status document so a
+previous run's report cannot be mistaken for current results. It carries two optional
+fields: `status` (`"failed"` or `"skipped"`) and `reason` (human-readable explanation,
+e.g. `"no changes detected"`). Both fields are absent from normal full reports.
+
 **Schema version history:**
 
 | Version | Changes |
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelCore.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelCore.java
@@ -32,9 +32,21 @@ public class ScalpelCore {
     private final Logger logger = LoggerFactory.getLogger(getClass());
     private final GitChangeDetector gitChangeDetector;
 
+    /**
+     * Human-readable reason when the last {@link #detectChanges} invocation deliberately skipped
+     * change detection (returned null for a non-error condition such as a disableOnBranch match),
+     * or null if the last invocation did not skip or skipped due to a failSafe bail-out.
+     * Set on each call; advisory only, so a stale value from a concurrent build is harmless.
+     */
+    private volatile String lastDetectionSkipReason;
+
     @Inject
     public ScalpelCore(GitChangeDetector gitChangeDetector) {
         this.gitChangeDetector = requireNonNull(gitChangeDetector, "gitChangeDetector");
+    }
+
+    public String getLastDetectionSkipReason() {
+        return lastDetectionSkipReason;
     }
 
     /**
@@ -48,11 +60,13 @@ public class ScalpelCore {
      */
     public ChangeDetectionResult detectChanges(Path reactorRoot, ScalpelConfiguration config, Set<String> allPomPaths)
             throws ScalpelException {
+        lastDetectionSkipReason = null;
         Repository repository;
         try {
             repository = openRepository(reactorRoot);
         } catch (RepositoryNotFoundException | IllegalArgumentException e) {
             logger.info("Scalpel: Not a git repository, building all modules");
+            lastDetectionSkipReason = "not a git repository";
             return null;
         } catch (IOException e) {
             return handleError(config, "Error opening git repository", e);
@@ -69,6 +83,7 @@ public class ScalpelCore {
                                     "Scalpel: Disabled because current branch '{}' matches pattern '{}'",
                                     currentBranch,
                                     pattern);
+                            lastDetectionSkipReason = "disabled by disableOnBranch";
                             return null;
                         }
                     }
@@ -88,6 +103,7 @@ public class ScalpelCore {
                     if (matchesSafely(baseBranchName, pattern, "disableOnBaseBranch")) {
                         logger.info(
                                 "Scalpel: Disabled because base branch '{}' matches pattern '{}'", baseBranch, pattern);
+                        lastDetectionSkipReason = "disabled by disableOnBaseBranch";
                         return null;
                     }
                 }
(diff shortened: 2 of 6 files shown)


gnodet (changes requested):
Well-structured fix for issue #89 with excellent TDD approach and thorough test coverage. However, two verified issues need attention before merge:

### 🔴 High: Null baseBranch crash in `writeStatusReport`

`writeStatusReport` calls `config.getBaseBranch()` and passes the result to `ScalpelReport.builder().baseBranch(...)`. When baseBranch is null (no CI env vars, no explicit config), `build()` throws `IllegalStateException("baseBranch is required")`. The method only catches `IOException`, so the exception propagates. Under `failSafe=true`, the catch block calls `writeFailedStatusReport`, triggering the same crash again — defeating failSafe in the exact scenario it exists for.

Tests miss this because `createSimpleSession` always hardcodes `baseBranch="base"`.

**Suggested fix:** Guard with a fallback and widen the catch:
```java
private void writeStatusReport(...) {
    try {
        String baseBranch = config.getBaseBranch();
        ScalpelReport report = ScalpelReport.builder()
                .baseBranch(baseBranch != null ? baseBranch : "(unconfigured)")
                // ...
                .build();
        report.writeToFile(reactorRoot, config.getReportFile());
    } catch (Exception e) {
        logger.warn("Could not overwrite report: {}", e.toString());
    }
}
```

### 🟡 Medium: Skip reason routing too narrow

`startsWith("disabled by")` only matches `"disabled by disableOnBranch"` and `"disabled by disableOnBaseBranch"`. The strings `"not a git repository"` and `"no base branch configured"` fall through to the else branch and are classified as `"failed"` with a generic reason, contradicting the PR description's stated intent to report them as `status:"skipped"`.

**Suggested fix:** Route on `skipReason != null` instead:
```java
if (skipReason != null) {
    writeStatusReport(config, reactorRoot, "skipped", skipReason);
} else {
    writeStatusReport(config, reactorRoot, "failed",
            "change detection did not run (see build log)");
}
```

The overall design is solid — `handleWriteFailure` extraction, the `lastDetectionSkipReason` volatile field, and the 7 new/updated tests demonstrate strong engineering.

### 📋 PR Metadata

| Milestone | _(none)_ | `0.3.11` |

paoloantinori:
Both review findings verified independently and fixed in 8153ba8.

1. Null baseBranch crash (HIGH): reproduced with a new test before fixing - a failSafe bail-out with no scalpel.baseBranch property and no CI env hit IllegalStateException: baseBranch is required from ScalpelReport.Builder.build(), thrown outside the IOException-only catch. writeStatusReport now uses a "(unconfigured)" placeholder when baseBranch is null, and the catch is widened to Exception with a WARN (status reports must never fail the build). The new test asserts the report is written with the placeholder, status "skipped", and the actual skip reason.

2. Skip-reason routing (MEDIUM): also reproduced - a "not a git repository" skip reason fell into the failed branch with a generic message. Routing now maps any non-null lastDetectionSkipReason to status "skipped" with that reason; only a null reason yields "failed" with "change detection did not run (see build log)". New test asserts "not a git repository" produces status "skipped" with the reason and not "failed"; the existing disabledByBranch test still passes unchanged.

Verification: ./mvnw -am -pl extension3 verify - BUILD SUCCESS (154 extension3 + 129 core tests, modernizer/jacoco/spotless gates included); ./mvnw -pl core test - BUILD SUCCESS (129 tests).

paoloantinori:
Follow-up on the review: both findings were valid and we've treated the HIGH one as a process failure on our side, not just a code bug - our pre-push review gates explicitly checked bail-out coverage and schema consistency, but never asked whether every configuration value consumed by the new code path had a null/unset test (every test fixture hardcoded a non-null baseBranch, which is exactly the blind spot you named). Our workflow now mandates, before any push: enumerate every config input the changed code reads and include a null-variant test for each, with that question baked into the review prompts. Thanks for catching what our gates should have.

gnodet (approved):
Both previous findings are fully addressed by commit 8153ba8. Changing verdict from REQUEST_CHANGES to APPROVE.

### Previous Findings — Resolved ✅

1. **~~Null baseBranch crash in writeStatusReport~~** — Fixed. `writeStatusReport` now applies a null guard (`baseBranch != null ? baseBranch : "(unconfigured)"`), and the catch was widened from `IOException` to `Exception`. A dedicated test confirms the placeholder is written when baseBranch is null.

2. **~~Skip reason routing too narrow~~** — Fixed. The condition was changed from `skipReason.startsWith("disabled by")` to `skipReason != null`. All skip reasons including "not a git repository" and "no base branch configured" are now correctly classified as `"skipped"`. Two new tests verify this.

### General Comments

- The fix commit is well-scoped: 11 lines of production changes and 78 lines of new tests.
- The comment "Status reports must never fail the build; that is their entire job" clearly documents the design decision.
- CI is green across all 12 matrix combinations.

### 📋 PR Metadata

| Milestone | _(none)_ | `0.3.11` |

---

PR #138, opened 2026-08-25
Fix #90: warn and report changed POMs that match no reactor project

### LINKED ISSUE #90: [ux] Changed POMs that match no reactor project are silently dropped
## Behavior

`PomChangeAnalyzer` skips a changed `pom.xml` whose path is not in `projectByPomPath` with only a debug log. That is correct for a deleted module, but a module present in `<modules>` yet absent from `session.getProjects()` has its POM change dropped entirely and silently. This happens when the module is profile-gated, or when `-pl` was used with `disableOnSelectedProjects` at its default of false.

The silence is the problem: an adopter sees Scalpel working, sees their POM changed, and the change simply does not participate in the analysis.

## Expected

- Unmatched changed POMs are counted and logged at WARN with their paths when non-zero.
- The count (and ideally the paths) appear in the JSON report, e.g. an `unmatchedPomPaths` field. Note: this would touch the report schema, so it should be coordinated with the schema work in #60 (optional fields without a version bump per that PR's policy).
- Optionally, a configuration flag to force a full build on unmatched POMs, defaulting to the safe behavior.

I have a branch implementing the WARN + report field with failing tests first; happy to rebase it onto #60 once that lands.

paoloantinori:
## Problem

As filed in #90: a changed `pom.xml` whose path matches no reactor project is dropped with only a debug log. That is correct for a deleted module, but a module present in `<modules>` yet absent from `session.getProjects()` (profile-gated, or `-pl` with `disableOnSelectedProjects=false` default) has its POM change silently ignored - the adopter sees Scalpel working, sees the POM changed, and the change simply does not participate in the analysis.

## Change

- `PomChangeAnalyzer` collects unmatched changed POM paths into `Result.unmatchedPomPaths` and emits a single WARN with the count and paths (naming the three causes: profile-gated, `-pl`-excluded, or removed module).
- `ScalpelReport` gains an optional `unmatchedPomPaths` array (additive, emitted only when non-empty, no version bump per the #60 policy); the participant wires it through; the checked-in `scalpel-report-v2.schema.json` documents it as optional.

## Verification

TDD: RED first (tests failing to compile against the missing API), then the implementation, then a review pass on the WARN wording. Full `core` (131) + `extension3` (148) suites green, including the golden-report comparison (field omitted when empty).

### DIFF
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
@@ -56,6 +56,7 @@ public final class ScalpelReport {
     private final List<String> changedProperties;
     private final List<String> changedManagedDependencies;
     private final List<String> changedManagedPlugins;
+    private final List<String> unmatchedPomPaths;
     private final List<AffectedModule> affectedModules;
     private final int excludedUpstreamCount;
 
@@ -67,6 +68,7 @@ public final class ScalpelReport {
             List<String> changedProperties,
             List<String> changedManagedDependencies,
             List<String> changedManagedPlugins,
+            List<String> unmatchedPomPaths,
             List<AffectedModule> affectedModules,
             int excludedUpstreamCount) {
         this.baseBranch = baseBranch;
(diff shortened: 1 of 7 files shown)


gnodet (approved):
Clean, well-scoped fix for issue #90 that surfaces previously-silent unmatched POM paths as a WARN log and an optional JSON report field. The implementation follows established codebase patterns, respects the forward-compatibility schema policy from PR #60, and includes thorough test coverage.

**Minor observation (non-blocking):**

The `null` check on `unmatchedPomPaths` in `toJson()` is defensive but unnecessary — the Builder initializes the field as `new ArrayList<>()` and the constructor is private. Other list fields don't have null checks in `toJson()`. Harmless, just a style inconsistency.

The WARN message wording is thoughtful — naming the three possible causes (profile-gated, `-pl`-excluded, or removed module) guides adopters toward the right diagnosis.

### 📋 PR Metadata

PR #139, opened 2026-08-25
Fix #91: report skipped modules with reasons

### LINKED ISSUE #91: [enh] Report should list skipped modules with reasons - the list a reviewer needs to trust a green trimmed build
## Problem

In trim mode the log says "Building 3 of 8 modules" and the JSON report lists AFFECTED modules. Nothing anywhere enumerates the SKIPPED set with the reason each was judged safe to skip. That is exactly the list a reviewer needs to trust a green trimmed build.

Anecdotally, four of five known adopters (Camel, Camel Quarkus, Camel Quarkus Examples, Apicurio) ship Scalpel wired up but switched off by default; only Quarkus runs it enabled. Reviewability of the skip decision is a plausible part of that.

Precedent: Bazel's `target-determinator` ships per-target reason output as a first-class feature; Nx renders an affected-project graph per CI run.

## Proposal

- A `skippedModules` array in the JSON report with path, groupId, artifactId and a reason for each.
- Reasons distinguishing at minimum `NOT_AFFECTED`, `EXCLUDED_BY_PATH_FILTER` and `TESTS_SKIPPED_ONLY`.
- Skipped counts in the summary log line; populated in trim, skip-tests and report modes.

Most of the data already exists in `TrimResult`, so this is largely a serialization task. I have a prototype branch with the `SkippedModule` serialization and `NOT_AFFECTED` populated; the other two reasons need a small design decision (path filtering operates on files, not modules). Happy to rebase and PR.

paoloantinori:
## Problem

As filed in #91: in trim mode the log says "Building N of M modules" and the report lists AFFECTED modules, but nothing enumerates the SKIPPED set with the reason each was judged safe to skip - exactly the list a reviewer needs to trust a green trimmed build.

## Change

- `skippedModules` array in the JSON report: path, groupId, artifactId and a reason per module, populated in trim and report modes. Computed as allProjects minus (directly affected + transitively affected + the FINAL build set after the includePaths scope filter, so scope-excluded downstream modules are listed, not invisible). Optional, additive, omitted when empty - no version bump per the #60 policy; schema updated with a `reason` enum.
- Only `NOT_AFFECTED` ships: it is the one reason reliably computable from existing sets. `EXCLUDED_BY_PATH_FILTER` needs the report-side filter to expose its decisions and `TESTS_SKIPPED_ONLY` describes modules that are built - both deferred as the issue itself suggests. Note: modules dropped by the includePaths scope currently carry NOT_AFFECTED; distinguishing them is the visible schema change of that follow-up.

**Deliberate behavior change**: trim mode now also writes the JSON report (that is what makes the skipped set reviewable where it matters). Two consequences, flagged for review: a report write failure can now fail a trim build (failSafe governs as elsewhere once #89 lands), and the zero-affected early-return paths write no report; happy to align either if you prefer.

## Verification

TDD across three commits (RED on the missing API and on the no-report-in-trim behavior, GREEN, then a review fix threading the filtered build set through with a dedicated includePaths test). Full suites green (core 131, extension3 149).

### DIFF
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
@@ -44,6 +44,12 @@ public final class ScalpelReport {
     public static final String REASON_TRANSITIVE_DEPENDENCY_TEST = "TRANSITIVE_DEPENDENCY_TEST";
     public static final String REASON_EXCLUDED_DOWNSTREAM = "EXCLUDED_DOWNSTREAM";
 
+    /**
+     * Skip reason for a reactor module that was left out of the build set because it is
+     * not affected by the changeset (not direct, transitive, upstream or downstream).
+     */
+    public static final String SKIP_REASON_NOT_AFFECTED = "NOT_AFFECTED";
+
     public static final String CATEGORY_DIRECT = "DIRECT";
     public static final String CATEGORY_UPSTREAM = "UPSTREAM";
     public static final String CATEGORY_DOWNSTREAM = "DOWNSTREAM";
@@ -57,6 +63,7 @@ public final class ScalpelReport {
     private final List<String> changedManagedDependencies;
     private final List<String> changedManagedPlugins;
     private final List<AffectedModule> affectedModules;
+    private final List<SkippedModule> skippedModules;
     private final int excludedUpstreamCount;
 
     private ScalpelReport(
@@ -68,6 +75,7 @@ public final class ScalpelReport {
             List<String> changedManagedDependencies,
             List<String> changedManagedPlugins,
             List<AffectedModule> affectedModules,
+            List<SkippedModule> skippedModules,
             int excludedUpstreamCount) {
         this.baseBranch = baseBranch;
         this.fullBuildTriggered = fullBuildTriggered;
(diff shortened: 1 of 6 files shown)


gnodet (approved):
Well-crafted enhancement that adds a `skippedModules` array to the JSON report, making trimmed builds reviewable by enumerating exactly which modules were skipped and why. The schema change is backward-compatible (additive, optional field, per #60), and the behavior change (trim mode now writes the report) is safe under the existing failSafe error handling. Test coverage is thorough via TDD.

**Minor observations (non-blocking):**

1. **Schema description** — The top-level `description` in `scalpel-report-v2.schema.json` still says "mode=report" but trim mode now produces the same report. Consider updating to reflect both modes.

2. **SkippedModule reason validation** — The `SkippedModule` constructor accepts any `String reason` without validation. Fine today with only `NOT_AFFECTED`, but when `EXCLUDED_BY_PATH_FILTER` and `TESTS_SKIPPED_ONLY` are added, a validation guard would prevent silent schema violations.

The `filteredBuildSet` threading through `AnalysisContext` is a clean approach, and the `trimMode_includePathsScopeExcludedModuleAppearsInSkippedModules` test is a particularly valuable edge case.

### 📋 PR Metadata

| Milestone | _(none)_ | `0.3.11` |

PR #140, opened 2026-08-25
Fix #111: integration-test the extension artifact users install

### LINKED ISSUE #111: [ci] The extension artifact users are told to install is never exercised by any test
## Problem

The README tells users to put `eu.maveniverse.maven.scalpel:extension` in their `.mvn/extensions.xml`, and Maven Central confirms that is the published user-facing coordinate. Camel, Camel Quarkus and Apicurio all reference `extension`; only Quarkus references `extension3` directly.

But the `extension` module contains exactly one file: `pom.xml`. No `src/`, and the integration tests depend on `extension3` directly. The artifact users install is never exercised by any test.

The module also declares a `maven-model-builder` dependency that `extension3` does not use and nothing in the repository references.

The intent is presumably the Maveniverse convention where `extension` is the stable user-facing coordinate and `extension3`/`extension4` are generation-specific implementations. That convention becomes load-bearing once Maven 4 support lands, so resolving it now is cheaper than later.

## Proposal

- An IT suite consuming the `extension` artifact through `.mvn/extensions.xml` exactly as a user would (smoke + one report-mode scenario).
- Remove or justify the dead `maven-model-builder` declaration.
- Document the extension/extension3 relationship in both POMs and the README.

paoloantinori:
## Problem

As filed in #111: the README tells users to put `eu.maveniverse.maven.scalpel:extension` in `.mvn/extensions.xml`, but the `extension` module has no `src/` and no test consumes that artifact - the coordinate every adopter except Quarkus actually installs was never exercised by any test.

## Change

New `it/extension-its` module (invoker wiring identical to `extension3-its`, dependency switched to the `extension` artifact), registered in `it/pom.xml`. Two scenarios consume the artifact exactly as a user would, via the invoker project's own `.mvn/extensions.xml` at `@project.version@`:

- **smoke**: two-module reactor; asserts Scalpel activation, the directly-affected line, and the correct Building N of M with the upstream module retained.
- **report-mode**: base-branch setup mirrored from extension3-its; asserts the report is written with version 2, `fullBuildTriggered:false`, and the SOURCE_CHANGE entry.

Control evidence: with the extensions.xml reference deliberately broken (nonexistent artifactId) the IT fails at extension resolution; restored, both scenarios pass. The extension module is a POM-only shim pulling in extension3 - noted in the issue - so beyond resolution there is no separate code path to cover; its `maven-model-builder` provided dependency remains unexercised by any code in the reactor (left in place, flagged for your call).

## Verification

Fresh full run of the new module passes (2/2 scenarios); extension3-its re-run unaffected; unit suites green.

### DIFF
--- a/it/extension-its/pom.xml
+++ b/it/extension-its/pom.xml
@@ -1,0 +1,98 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) Maveniverse Org.
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v2.0
+    which accompanies this distribution, and is available at
+    https://www.eclipse.org/legal/epl-v20.html
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>eu.maveniverse.maven.scalpel.it</groupId>
+    <artifactId>it</artifactId>
+    <version>${nisse.jgit.dynamicVersion}</version>
+  </parent>
+
+  <artifactId>extension-its</artifactId>
+  <packaging>pom</packaging>
+
+  <name>${project.groupId}:${project.artifactId}</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>eu.maveniverse.maven.scalpel</groupId>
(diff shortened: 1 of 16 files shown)


gnodet (approved):
Good integration test coverage for the extension artifact. The test validates the core trimming workflow and verifies artifact packaging.

1. **Smoke test doesn't verify trimming** — The integration test validates that the extension builds and packages correctly, but doesn't assert that the trimming behavior actually occurred (e.g., checking that excluded dependencies are absent from the final artifact). Consider adding a verification step.

2. **setup.groovy duplication** — The `setup.groovy` file shares boilerplate with other integration test setups. Consider extracting common setup logic into a shared script to reduce maintenance burden.

### 📋 PR Metadata

| Category | _(unlabeled)_ | `test` |

---

PR #141, opened 2026-08-25
Fix #109: integration tests for includePaths, disableOnBaseBranch, failSafe=false, CI base detection

### LINKED ISSUE #109: [ci] Shipped configuration paths with no integration coverage (includePaths, disableOnBaseBranch, CI auto-detection, failSafe=false, pluginManagement)
## Problem

Behaviors shipped and documented but never exercised end to end:

- `includePaths` (the newest feature) filters four separate code paths: the affected set, the trim build set, skip-tests assignment and the report's downstream section. No IT at all.
- `disableOnBaseBranch` has no behavioral test of any kind, including the remote-prefix stripping, which has a latent bug: it strips at the first slash, so a local base branch named `release/1.0` without a remote prefix becomes `1.0`.
- CI base-branch auto-detection (GITHUB_BASE_REF / CI_MERGE_REQUEST_TARGET_BRANCH_NAME / CHANGE_TARGET plus precedence) has zero tests despite being the primary production path.
- Managed-plugin transitive impact is unit-tested but no IT mutates `pluginManagement`.
- `failSafe=false` is never proven to fail a real build: no IT sets `invoker.buildResult = failure`.
- Binary-file skip, `maxResourceFileSize`, `untracked`, custom `reportFile`, non-default `scalpel.head`: no ITs.

Also worth fixing while in there: only one of the ~23 ITs runs `package` rather than `validate` (the only one that could catch a real compile/surefire failure), and most assert on log substrings only.

## Proposal

ITs for each of the above, extending the existing `it/extension3-its` harness; more ITs running `package`; a `fail-safe-disabled` IT asserting the build actually fails.

paoloantinori:
## Problem

As filed in #109: several configuration paths shipped and documented but never exercised end to end - `includePaths` (the newest feature, four code paths, zero ITs), `disableOnBaseBranch`, `failSafe=false` (no IT in the suite asserted a failure outcome), and CI base-branch auto-detection (the primary production path).

## Change

Four new scenarios in `it/extension3-its/src/it/`:

- **include-paths**: runs `package` (one of the few ITs doing more than validate) with `scalpel.includePaths=module-a/**`; asserts module-a directly affected and built, downstream module-b excluded from the Building line. Control run without the filter builds all 3 - the IT detects the difference.
- **disable-on-base-branch**: pattern matching the configured base branch; asserts the disable log and that no affected-set computation runs.
- **fail-safe-disabled**: `invoker.buildResult = failure` with a nonexistent baseBranch and `failSafe=false`; asserts the build fails with the Scalpel error line (the first IT in the suite asserting a failure outcome).
- **ci-base-detection**: `GITHUB_BASE_REF` propagated via `invoker.environmentVariables`, no explicit baseBranch; a bare origin with `origin/base` pushed. Detection is proven by elimination: trimming happened, which requires a resolved merge base, and the "No base branch configured" log is absent.

## Verification

Fresh targeted run: 4/4 pass; full extension3-its suite (28 ITs) green; unit suites green. Two harness findings baked into the scenarios: commas in invoker.properties goal lines get mangled (single patterns used), and a MavenExecutionException from the extension aborts before the BUILD FAILURE footer, so the failure IT asserts on the `[ERROR] Scalpel:` line.

### DIFF
--- a/it/extension3-its/src/it/ci-base-detection/.mvn/extensions.xml
+++ b/it/extension3-its/src/it/ci-base-detection/.mvn/extensions.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<extensions>
+    <extension>
+        <groupId>eu.maveniverse.maven.scalpel</groupId>
+        <artifactId>extension3</artifactId>
+        <version>@project.version@</version>
+    </extension>
+</extensions>
--- a/it/extension3-its/src/it/ci-base-detection/invoker.properties
+++ b/it/extension3-its/src/it/ci-base-detection/invoker.properties
@@ -1,0 +1,2 @@
+invoker.goals = validate
+invoker.environmentVariables.GITHUB_BASE_REF = base
--- a/it/extension3-its/src/it/ci-base-detection/module-a/pom.xml
+++ b/it/extension3-its/src/it/ci-base-detection/module-a/pom.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>eu.maveniverse.maven.scalpel.it.cibasedetection</groupId>
+        <artifactId>ci-base-detection</artifactId>
+        <version>0.1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>module-a</artifactId>
+
+    <name>${project.groupId}:${project.artifactId}</name>
+</project>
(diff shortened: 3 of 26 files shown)


gnodet (approved):
A well-structured, test-only PR adding four new integration tests that fill significant coverage gaps for existing Scalpel configuration paths (includePaths, disableOnBaseBranch, failSafe=false, CI base detection). All assertions align with the source code's actual log output, the ITs follow established project patterns, and the scenarios are correctly designed.

1. **New `invoker.environmentVariables` pattern** — `ci-base-detection/invoker.properties`: This is the first IT in the suite to use `invoker.environmentVariables.*`. While this works correctly (requires Maven Invoker Plugin >= 3.5.0, which the parent POM provides), it introduces a new pattern worth noting for future contributors.

### General Comments

- The `include-paths` IT runs `package` rather than `validate`, exercising actual compilation and module-dependency resolution under trimming — valuable since most existing ITs only run `validate`.
- The `fail-safe-disabled` IT is the first scenario in the entire suite to assert a build failure outcome (`invoker.buildResult = failure`), filling a notable gap.
- The `ci-base-detection` setup with a bare remote and `GITHUB_BASE_REF` propagation is a faithful reproduction of the real CI detection path, and the "proof by elimination" verification approach is sound.
- All four verify.groovy files correctly match substring patterns from the actual SLF4J-formatted log output. The assertions are well-anchored to the source.
- The PR covers 4 of the 6+ items listed in issue #109. The remaining items could be follow-up work — the incremental approach is appropriate.

### 📋 PR Metadata

| Category | _(unlabeled)_ | `test` |

---

PR #142, opened 2026-08-25
Fix #93: explain mode with per-module decision evidence

### LINKED ISSUE #93: [enh] Explain mode: per-module decision evidence (which file, which property, which GA triggered the decision)
## Problem

Today, answering "why did Scalpel skip my module?" or "why did it build everything?" requires reading the source. The `reasons` array in the report gives the category but not the specific input that triggered it.

## Proposal

`-Dscalpel.explain=true`, valid in every mode, dumping for each module its decision (BUILD or SKIP) plus the specific evidence: which changed file, which property, which managed dependency GA, which plugin, or which module it is upstream of. Each SKIP decision names what was checked and found not to apply. Output as a log section and as a field in the JSON report.

Precedent: `target-determinator -verbose` explains why each target is selected, and it is the feature that makes the Bazel equivalent debuggable rather than magical. Would also be the natural first diagnostic step for a future troubleshooting guide.

paoloantinori:
## Problem

As filed in #93: today answering "why did Scalpel skip/build this module?" requires reading the source. The report's `reasons` array gives the category (SOURCE_CHANGE etc.) but not the specific input that triggered it.

## Change

`-Dscalpel.explain=true`, valid in every mode:

- Report module entries gain an optional `evidence` array naming the specific input: the triggering changed file(s), `property <name>`, `managed dep <ga>`, `managed plugin <ga>`, `filtered resources referencing changed properties`, `downstream of <ga>`, `upstream of <ga>`, `forced by forceBuildModules pattern <p>`. Emitted only when explain=true; with explain=false the report is byte-identical (the golden-file test passes unchanged).
- A per-module INFO line per decision in every mode (`Scalpel explain: BUILD module-x because: ...`).
- Schema gains the optional evidence array (no version bump, per the #60 policy).
- Evidence collection is gated on the flag: no string building or map population runs on builds that did not ask for it (a review pass caught the initial version collecting unconditionally).

**Honest scope notes**: parent-self-affected entries currently carry no specific input (the parent-diff block computes six independent booleans; naming which one fired needs a small restructuring of that block); BOM-importer children get the same evidence as inheritance children; multi-hop chains name the immediate neighbor, not the originating module. All are natural follow-ups if you want them.

## Verification

TDD (RED on the missing API, GREEN, plus the gating review commit); full suites green (core 134, extension3 149) including the golden-file byte-identity test.

### DIFF
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
@@ -27,6 +27,7 @@ public final class ScalpelConfiguration {
     public static final String FULL_BUILD_TRIGGERS = PREFIX + "fullBuildTriggers";
     public static final String FAIL_SAFE = PREFIX + "failSafe";
     public static final String MODE = PREFIX + "mode";
+    public static final String EXPLAIN = PREFIX + "explain";
 
     public static final String DISABLE_ON_BRANCH = PREFIX + "disableOnBranch";
     public static final String DISABLE_ON_BASE_BRANCH = PREFIX + "disableOnBaseBranch";
@@ -69,6 +70,7 @@ public final class ScalpelConfiguration {
             FULL_BUILD_TRIGGERS,
             FAIL_SAFE,
             MODE,
+            EXPLAIN,
             DISABLE_ON_BRANCH,
             DISABLE_ON_BASE_BRANCH,
             EXCLUDE_PATHS,
@@ -112,6 +114,7 @@ public final class ScalpelConfiguration {
     private final String impactedLog;
     private final boolean failSafe;
     private final String mode;
+    private final boolean explain;
     private final String reportFile;
     private final long maxResourceFileSize;
     private final List<String> warnings;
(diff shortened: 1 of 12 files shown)


gnodet (approved):
Well-implemented opt-in explain mode (`-Dscalpel.explain=true`) that surfaces per-module decision evidence in both the JSON report and INFO logs. Evidence collection is properly gated behind the flag in production code paths, the schema change is additive and non-breaking, and backward compatibility is preserved through constructor overloads.

1. **Fully-qualified type name** — `TrimResult.java` line ~30: The `buildReasons` field uses `java.util.List<String>` in three places, but `java.util.List` is already imported. Using the short form `List<String>` would be consistent with the rest of the file.

2. **Missing explain-mode test for non-report modes** — `ScalpelLifecycleParticipantTest.java` line ~3090: All three explain-mode integration tests use `report` mode. The PR states explain mode is "valid in every mode," so a test for at least `trim` or `skip-tests` mode would strengthen confidence. Low priority since the evidence-building logic is shared.

### General Comments

- The evidence collection gating behind `config.isExplain()` is correctly threaded through all production code paths, ensuring zero overhead for the common case.
- The JSON schema change is additive and optional — existing report consumers will not break.
- The "Honest scope notes" section in the PR description transparently documents known gaps, making follow-up work easy to scope. Excellent practice.
- The introduction of `TransitiveMatch` and `ChangedDependencyMatch` to carry evidence is a clean pattern that avoids polluting existing return types.
- The `findChangedPlugin` refactor (boolean → matched GA) makes the method more useful without changing its caller's contract.

### 📋 PR Metadata

---

Issue #143, opened 2026-08-25
Scalpel should gracefully degrade when git merge-base is unavailable


gnodet:
## Problem

Scalpel is loaded as a Maven core extension via `.mvn/extensions.xml`, which means it initializes for **every** `./mvnw` invocation — including trivial utility goals like `help:evaluate -N` that have nothing to do with reactor trimming.

When git state is incomplete (e.g., shallow clone without a reachable merge-base), Scalpel fails hard and causes Maven to **exit with code 1** instead of gracefully degrading to a no-op.

## Reproduction

This was observed in Apache Camel CI ([failed run](https://github.com/apache/camel/actions/runs/32765151716)):

1. Camel's CI workflow has a step that fetches the base branch to establish the merge-base for Scalpel
2. When using `gh run rerun --failed`, previously-succeeded steps are skipped — including the git fetch step
3. A subsequent step runs `./mvnw help:evaluate -Dexpression=... -q -DforceStdout -N -f parent/pom.xml`
4. Scalpel loads during this invocation, fails to find the merge-base, and crashes Maven

The workaround is to pass `-Dscalpel.enabled=false` to invocations that don't need change detection ([apache/camel#25662](https://github.com/apache/camel/pull/25662)), but ideally Scalpel itself should handle this gracefully.

## Expected behavior

When Scalpel cannot determine the merge-base (missing remote ref, shallow clone, detached HEAD without history, etc.), it should:

1. Log a warning explaining why change detection is unavailable
2. Fall back to a no-op (full reactor, no trimming)
3. **Not** cause Maven to fail

This is especially important for:
- CI re-run scenarios (as described above)
- Simple utility goals (`help:evaluate`, `help:effective-pom`, `versions:*`) where reactor trimming is irrelevant
- Shallow clones (`actions/checkout` with default `depth: 1`)

## Environment

- Scalpel version: 0.3.10 (`eu.maveniverse.maven.scalpel:extension3`)
- Maven: 3.9.x (via `./mvnw`)
- Java: 25 (where the failure was observed, but the issue is JDK-independent)

gnodet:
Closing this — further investigation showed Scalpel was not the cause.

[...]

PR #145, opened 2026-08-25
Fix #131: use effective POM models to detect managed dependency changes

### LINKED ISSUE #131: New managed dependencies in BOM should not trigger full downstream rebuild
## Problem

When a PR adds **new** entries to a BOM's `<dependencyManagement>` section (e.g. `camel-bom/pom.xml`), Scalpel reports all modules that import the BOM as affected — even though **no existing module actually depends on the newly added artifacts**.

In Apache Camel's CI ([PR #22158](https://github.com/apache/camel/pull/22158)), adding 4 new `camel-state-store-*` modules to the BOM caused Scalpel to report **620 affected modules** (597 would be tested), triggering a full ~1.5h build instead of the ~5 min incremental build that would cover only the new modules.

```
Scalpel detected 620 affected modules (597 would be tested)
  Direct: 14, Downstream tested: 583, Downstream skipped: 23
  Changed managed deps: org.apache.camel:camel-state-store, org.apache.camel:camel-state-store-caffeine,
                        org.apache.camel:camel-state-store-infinispan, org.apache.camel:camel-state-store-redis
```

## Expected behavior

Scalpel should distinguish between:

1. **Changed managed dep that is actually consumed** — an existing module's resolved dependency tree changes → must rebuild downstream
2. **New managed dep that nobody references** — no module resolves the new GAV → safe to skip downstream expansion

Adding a managed dependency that no module uses is a no-op for every existing module's effective dependency tree.

## Suggested approach

During the `mvn validate` model comparison, Scalpel already walks the effective POM for each module. For each `changedManagedDependency`, it could check whether any module in the reactor actually resolves that GAV (directly or transitively). If no module references the new managed dep, it should not be reported as a change that affects downstream modules.

This would keep the conservative behavior for genuine version bumps of existing managed deps (which do affect consumers), while avoiding false positives for purely additive BOM changes.

## Environment

- Scalpel mode: `report`
- Project: Apache Camel (~620 modules)
- CI run: https://github.com/apache/camel/actions/runs/32765151716

gnodet:
## Summary

When a BOM adds new `<dependencyManagement>` entries that no existing module actually depends on, Scalpel no longer reports all BOM importers as affected.

### Effective POM model comparison
- Refactored `PomChangeAnalyzer` to compare effective (resolved) models instead of raw XML — uses Maven's `ModelBuilder` to resolve inheritance, property interpolation, and BOM flattening
- Replaced the manual property-chasing approach (`augmentWithPropertyReferences`, `referencesAnyProperty`, etc.) — effective models already have properties interpolated
- Removed `identifyNewOnlyManagedDeps` — the existing `computeTransitivelyAffected` step already resolves each module's dependency tree and naturally filters out managed dep GAs that no module uses

### Real parent/BOM resolution via ProjectModelResolver
- Replaced the offline `NoOpModelResolver` with Maven's `ProjectModelResolver` backed by the local repository cache (`~/.m2/repository`)
- External parents and BOM imports now resolve fully — no more partial effective models from unresolvable parents
- Plugin processing stays disabled (`setProcessPlugins=false`) because lifecycle defaults depend on reactor state that temp-dir model building lacks

### Effective plugin comparison per child
- Added `hasEffectiveChanges()` — compares each child's old vs new effective plugins to catch property-driven plugin configuration changes (source versions, execution phases, etc.)
- GA-level matching alone misses these; the effective plugin comparison fills that gap
- Property text search kept as fallback for filtered resources (`${prop}` in `application.properties` etc.)

### Code quality
- Introduced `AnalysisContext` inner class to reduce method parameter counts (SonarCloud S107)
- Extracted helper methods to reduce cognitive complexity (S3776)
- Added debug log guards for string concatenation (S2629)
- Suppressed false positive on `Files.createTempDirectory` (S5443)
- Secure temp directory with owner-only POSIX permissions
- 0 new SonarCloud issues

Fixes #131

## Test plan

- [x] All 145 tests pass (61 unit + 62 integration + 22 other)
- [x] Integration test `bom-import-new-dep` — BOM adds both a version bump and a brand-new unused managed dep; verifies unused dep does not trigger downstream rebuild
- [x] Existing `bom-import` IT still passes (no regression on version bump detection)
- [x] Property interpolation in effective models verified (`reportMode_camelLike_kafkaVersionInChildDepsOnly`)
- [x] Profile-aware change detection verified (`profile-aware` IT)
- [x] Resource filtering verified (`resource-filtering` IT)
- [x] CI green on all 9 matrix combinations (ubuntu/macOS/windows × JDK 17/21/25)
- [x] SonarCloud passes with 0 new issues

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Improved Maven change analysis using effective POM information, including inherited properties, dependency management, plugins, profiles, and BOM imports.
  - More accurately identifies direct, transitive, and module-level impacts from parent and managed dependency changes.
  - Supports repository-aware analysis of historical project configurations.

- **Bug Fixes**
  - Prevents unused managed dependencies from affecting unrelated modules or appearing as changed.
  - Reduces false positives in plugin and dependency comparisons.
  - Normalizes default dependency scopes and types.

- **Tests**
  - Added coverage for property interpolation, inheritance, BOM updates, profiles, and transitive impact reporting.


extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -408,11 +413,6 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
             }
 
         } catch (ScalpelException e) {
-            if (config.isFailSafe()) {
-                logger.warn("Scalpel: {}, building all modules", e.getMessage());
-                logger.debug("ScalpelException details", e);
-                return;
-            }
             throw new MavenExecutionException("Scalpel: " + e.getMessage(), e);
```
gnodet, line 508: 🔴 **Regression of #143**: The `failSafe` guard for `ScalpelException` was removed here. PR #144 (commit 1b4d6c0) added this guard so that shallow clones and incomplete git history degrade gracefully instead of crashing the build:

```java
if (config.isFailSafe()) {
    logger.warn("Scalpel: {}, building all modules", e.getMessage());
    logger.debug("ScalpelException details", e);
    return;
}
```

Since `ScalpelException extends Exception`, this specific `catch (ScalpelException e)` handler fires **before** the general `catch (Exception e)` handler — so the failSafe guard in the general handler never gets a chance to intercept `ScalpelException`.

The two associated tests (`scalpelException_failSafeTrue_buildsAll` and `scalpelException_failSafeDisabled_throws`) were also deleted. This looks like an accidental merge/rebase artifact since both this PR and #144 modified the same catch block.

**Fix**: Restore the failSafe guard in the `catch (ScalpelException e)` block and restore the two test methods.

gnodet (commented):
## Review Summary

Excellent fix for #131 — the effective-model-comparison approach is sound and eliminates the fragile manual property-chasing logic. The integration test (`bom-import-new-dep`) validates the exact scenario from the issue. However, the PR inadvertently regresses #143 by removing the failSafe guard for `ScalpelException` that PR #144 just added.

### Verified Findings

| # | Severity | File | Finding | Verdict |
|---|----------|------|---------|---------|
| 1 | 🔴 high | `ScalpelLifecycleParticipant.java` | failSafe guard for ScalpelException removed — regresses #143 | ✅ confirmed |
| 2 | 🟡 low | `PomChangeAnalyzer.java` | `AnalysisContext` non-accumulator fields could be `final` | ✅ confirmed |

### 📋 PR Metadata

| Category | `bug` | `bug` + `enhancement` (effective-model architecture is an enhancement alongside the bug fix) |
| Labels | `bug`, `java` | + `enhancement` |

gnodet:
Addressed review feedback from @gnodet:

**ScalpelException failSafe guard regression** — Restored the `failSafe` check in the `catch (ScalpelException e)` block and re-added both associated tests (`scalpelException_failSafeTrue_buildsAll` and `scalpelException_failSafeDisabled_throws`). This was an accidental rebase artifact since both this PR and #144 modified the same catch block.

Fixed in b0e592e.

gnodet (commented):
## Re-review: ✅ Regression Fixed

The fix commit `b0e592e` properly restores the `catch(ScalpelException)` failSafe guard that was lost during rebase. Both the production code and companion tests are restored.

### Previous Finding Resolution

| # | Severity | Finding | Status |
|---|----------|---------|--------|
| 1 | 🔴 high | failSafe guard for ScalpelException removed (regresses #143) | ✅ **Fixed** — guard restored with exact PR #144 pattern |
| 2 | 🟡 low | AnalysisContext fields could be `final` | ℹ️ Pre-existing, non-blocking |

### What the fix does

- Restores `config.isFailSafe()` check in `catch(ScalpelException e)` before rethrowing
- Restores `scalpelException_failSafeTrue_buildsAll` test (graceful degradation)
- Restores `scalpelException_failSafeDisabled_throws` test (exception propagation)
- Follow-up commit `91eebd1` is clean dead-code removal (`OFFLINE_MODEL_RESOLVER` + convenience overloads)

LGTM 👍 (posted as COMMENT because GitHub does not allow approving your own PR)

### 📋 PR Metadata

gnodet (commented):
## Re-review: New commits look good ✅

Reviewed the 3 new commits since last approval:

| Commit | Summary | Assessment |
|--------|---------|------------|
| `91eebd1` | Remove OFFLINE_MODEL_RESOLVER + convenience overloads | ✅ Clean dead-code removal |
| `5edafe2` | Set parent before building effective model in test fixtures | ✅ Test alignment fix |
| `7c1d9f0` | Compare effective dependency tree instead of managed dep declarations | ✅ Core algorithm upgrade — well-designed |

### Highlights

- The layered detection approach (effective model comparison → GA matching → property text search → filtered resources) elegantly solves #131 while maintaining safety against false negatives
- `diffDependencies` vs `diffEffectiveDependencies` split is well-motivated — managed deps track only modifications/removals, effective deps track all changes
- `normalizeScope`/`normalizeType` helpers correctly prevent false diffs between raw and effective models
- failSafe guard confirmed intact

### Minor nit (non-blocking)

- `PomChangeAnalyzer.java:315-324`: When old effective model is null, the conservative bail-out correctly marks all dependents as affected but doesn't record explain-mode evidence (unlike the analogous exception handler at lines 327-338). Affected modules from this path will appear without an evidence string in explain/report mode.

LGTM 👍 (posted as COMMENT because GitHub does not allow approving your own PR)

gnodet (commented):
## Re-review (new commits since last approval)

The 8 new commits since the previous approval are a well-executed series of refinements that move POM change analysis from raw XML diffing to effective model comparison, eliminating false positives tested against Apache Camel (~700 modules). CI is fully green, test coverage is thorough, and the design is sound.

### Strengths

- **Symmetric effective model building** — using the same `ModelBuilder` instance for both old and new states is the right solution to eliminate false positives from model-builder asymmetry. The `ReactorFirstModelResolver` correctly resolves parents and BOMs from the temp directory containing old POM content.
- **Proper normalization** — `normalizeScope` (null → "compile") and `normalizeType` (null → "jar") correctly align with Maven's default behavior and prevent false positive diffs.
- **Well-motivated separation** — `diffDependencies` (modifications/removals only for managed deps) vs. `diffEffectiveDependencies` (all changes for effective dep trees) correctly handles the #131 requirement.
- **Thorough integration test** — `bom-import-new-dep` validates the end-to-end scenario with a real Maven build.
- **Clean temp directory handling** — `createSecureTempDirectory` with POSIX owner-only permissions, `deleteRecursive` in finally block.

### Observations (non-blocking)

- `diffManagedPluginVersions` only reports modifications and removals, not additions (unlike `diffEffectiveDependencies` which reports all changes). This asymmetry means a parent adding a new `<build><plugin>` that children inherit will not flag those children as affected. The commit messages document this as intentional (configuration diffs covered by POM changeset), and the scenario is rare, but worth noting for future reference.
- `AnalysisContext` lookup-map fields are non-final (accepted in previous review cycle — style only).
- `ReactorModelResolver` and test helper methods are duplicated between `PomChangeAnalyzerTest` and `ScalpelLifecycleParticipantTest` — could be extracted to a shared test utility.

### 📋 PR Metadata

_Note: This PR would receive APPROVE but the reviewer cannot approve their own PR._

---

gnodet (commented):
## Re-review (2 new commits)

Both new commits are well-reasoned refinements:

- **589ff517** correctly replaces indirect GA-set-based transitive detection with direct effective-model/dependency-tree comparison, eliminating the intermediate abstraction that was both incomplete and imprecise.
- **9d8a59f2** adds a necessary second-pass fixed-point propagation through reactor dependencies to cover the limitation that Maven's reactor always provides current artifact versions during "old" tree resolution.

### Observations (non-blocking)

- **Test coverage for reactor propagation** — The second-pass `while(propagated)` loop (lines 655-703) has no dedicated test. The existing `reportMode_includesTransitivelyAffectedModules` test catches changes via first-pass tree comparison. A test exercising the specific scenario where old/new trees are identical because the reactor provides the same artifact version on both sides would strengthen confidence.
- The `findChangedDependencyInTree` scope defaulting for removed dependencies is conservatively correct (defaults to "compile" scope, ensuring tests run).
- The `applySkipTests` path correctly omits the second-pass propagation since `ReactorTrimmer.computeBuildSet` already handles reactor dependents through the trim result's downstream set.

### 📋 PR Metadata

_Note: This PR would receive APPROVE but the reviewer cannot approve their own PR._

---

gnodet (commented):
## Re-review (2 new commits: docs + property text-search replacement)

Both commits are solid:

- **31b641b7** adds excellent algorithm documentation — thorough, well-structured, and accurately describes the multi-phase analysis. The "Design decisions" section explaining why managed dependency GA sets are no longer used is particularly helpful.
- **8bf08da3** correctly replaces the crude property text-search fallback with principled effective model comparison. The approach of checking effective values of properties defined in the child's raw model avoids cascading false positives while still catching property-driven changes. The addition of raw-model managed dep and managed plugin comparison rounds this out well.

### Items to consider

- **Dead config code** — `maxResourceFileSize` in `ScalpelConfiguration` is now unused (field, constant, getter, parsing, validation, toString all remain but `PomChangeAnalyzer` no longer accepts it). Worth a cleanup commit.
- **Size limit change** — `checkFileForPropertyRefs` hardcodes 1MB (was configurable 10MB default). Also changed behavior: oversized text files are now silently skipped instead of conservatively marked as affected.
- **Removed test coverage** — 5 resource-filtering tests removed (-425 lines) with no replacements for the new effective property comparison. Existing tests cover effective dep comparison but not the new raw-model property check path.
- **Edge case** — When effective models can't be built (null), `hasEffectiveChanges` returns false and only the filtered-resource scan remains as safety net. The old property text-search provided additional coverage here.

### 📋 PR Metadata

_Note: This PR would receive APPROVE but the reviewer cannot approve their own PR._

---

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
paoloantinori: Verified on current main (2a09b78): the failSafe guard is present at ScalpelLifecycleParticipant.java:504 with exactly this shape, and the two scalpelException_failSafe* tests cover both branches, so the regression is not live. Marking the thread resolved - thanks for flagging it.

PR #151, opened 2026-08-26
Fix #84: validate module paths written to the impacted log

### LINKED ISSUE #84: [security] Impacted log writes PR-controlled module paths verbatim: unsafe-by-default contract for CI scripts
## Behavior

The impacted log writes module directory names one per line, verbatim, with no escaping or character validation. The line content is fully controlled by a PR author, because module directory names come from the reactor, which comes from the PR.

The README advertises the file "for CI scripts that need a simple flat file (e.g. for GitHub Actions matrix filtering)", and the natural implementations of that are unsafe, e.g. iterating over `$(cat scalpel-impacted.log)` or piping `jq` output into `xargs`. Spaces, semicolons, pipes, backticks, command substitution and a leading dash are all legal in Linux directory names, so a PR adding a module under a directory named with shell metacharacters gets that string executed with the runner's secrets in scope.

This is a dangerous-by-default **output contract** rather than a bug inside the analysis: the consumer writes the injectable shell. But Scalpel is the component that can make it safe cheaply.

## Expected

- Module paths containing characters outside a documented safe set are skipped with WARN; control characters and newlines always rejected (a newline would additionally forge entries and break the one-path-per-line contract).
- Optionally a NUL-delimited output option for `xargs -0` consumers.
- The README impacted-log section carries a warning and shows the safe consumption pattern (quoted `read` loop); same warning for `jq`-based consumption of the report's `path` field.

## Suggested fix

I can PR this with failing tests first (module path with shell metacharacters must not reach the output file unescaped). Happy to proceed if welcome.

paoloantinori:
## Problem

As filed in #84: the impacted log writes module directory names verbatim, one per line, with no escaping or character validation. The line content is controlled by a PR author (module directory names come from the reactor, which comes from the PR), and the README advertises the file "for CI scripts (e.g. for GitHub Actions matrix filtering)" - whose natural consumption patterns (`$(cat ...)`, xargs) execute shell metacharacters. Spaces, semicolons, pipes, backticks, command substitution and a leading dash are all legal in directory names.

This is a dangerous-by-default **output contract**: the consumer writes the injectable shell; Scalpel is the component that can make it safe cheaply.

## Change

- Every line is validated against a conservative allowlist (`A-Za-z0-9 - _ . /`, no leading dash). Control characters and newlines are rejected by construction, closing both the injection and the entry-forgery (newline in a module name splitting into extra entries) vectors. Paths outside the set are skipped with a WARN naming the module and the escaped path.
- The reactor root (relative path is the empty string) is written as `.` instead of the old blank line, so an affected root stays visible to consumers.
- README: the impacted-log section carries a warning plus the safe consumption patterns (quoted read loop, `xargs -d '\n'` GNU note), and the jq pattern for the report's `path` field filters to the same character set first - without the filter, `jq -r` emits embedded newlines raw and a read loop splits them into forged entries (empirically verified during review).

**Deliberate behavior changes**: unsafe paths no longer reach the file (WARN instead); the root module appears as `.` instead of a blank line; if every entry is rejected the log is an empty file with a console WARN as the only signal.

## Verification

TDD: RED first - `module;rm`, `module$(id)` written verbatim and a directory named with an embedded newline forging a second entry (tests fail on main exactly as predicted); then the fix, then the review pass (root-as-`.`, jq filter). Full suites green (extension3 166, core 143). NULL-input coverage per the config values the path reads (unset, blank, null, empty).

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,26 @@ module-b/sub-module
 This works in all modes (trim, skip-tests, report) and is written alongside the JSON report,
 not as a replacement. It contains only directly affected modules (not upstream/downstream).
 
+**Safety.** Module directory names are chosen by whoever authors the repository (or pull request),
+so the log is written defensively: a path is only written when it consists solely of letters,
+digits, `-`, `_`, `.` and `/`, and does not start with `-`. A path containing anything else
+(spaces, quotes, `$`, backticks, `;`, `|`, glob characters, backslashes, newlines, control
+characters) is skipped with a WARN instead of written. Even so, consume the file defensively:
+unquoted expansion word-splits and glob-expands file contents.
+
+```bash
+# Safe: quoted read loop, one module per line
+while IFS= read -r module; do
+  build_module "$module"
+done < target/scalpel-impacted.log
+
+# Safe (GNU xargs): explicit newline delimiter, no word splitting
+xargs -d '\n' build_module < target/scalpel-impacted.log
+
+# Unsafe: unquoted command substitution word-splits and glob-expands the contents
+build_all $(cat target/scalpel-impacted.log)
+```
+
 #### Report Format
 
 The report follows a versioned JSON schema. A [JSON Schema](core/src/main/resources/eu/maveniverse/maven/scalpel/core/scalpel-report-v2.schema.json)
@@ -155,6 +175,18 @@ is published alongside the code. The current version is **2**.
     }
   ]
 }
+```
+
+**Consuming `path` values in scripts:** like the impacted log, the `path` fields of
+`affectedModules` entries carry module directory names chosen by the repository (or pull request)
+author. The values are JSON-encoded, so newlines and quotes cannot corrupt the document, but they
+are not restricted to a safe character set. Keep them quoted end-to-end when feeding them to shell
+commands:
+
+```bash
+# Safe: jq emits one path per line; consume it with a quoted read loop
+jq -r '.affectedModules[].path' target/scalpel-report.json |
+  while IFS= read -r module; do build_module "$module"; done
 ```
 
 **Status-only reports:** when analysis does not complete (failSafe bail-out, unexpected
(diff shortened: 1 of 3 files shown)


gnodet (approved):
Strong security fix. The output contract is dangerous-by-default (CI scripts `$(cat ...)` the log), and this closes the injection vector properly.

**Implementation review:**

- `isSafeImpactedLogPath`: Correct conservative allowlist. Character-by-character scan, no regex overhead. Rejects null, empty, leading dash (option injection). The allowed set `[A-Za-z0-9._/-]` covers all reasonable module directory names while blocking every shell metacharacter class (`;`, `|`, `$()`, backticks, quotes, glob chars, whitespace, control chars, newlines).
- `escapeControlChars`: Properly defends the WARN message itself against log forging — a rejected path with embedded newlines could otherwise forge extra log lines in the skip warning.
- Empty → `"."` mapping for the reactor root: correct, `"."` passes validation, and consumers see a visible entry instead of a blank or skipped line.
- `writeImpactedLog` flow: validate-then-add loop is clean. Files.write with the filtered list is atomic enough (writes all lines at once).
- README: safe consumption patterns (quoted read loop, `xargs -d '\n'`) and the jq filter with `select(test(...))` are well-documented and consistent with the server-side character set.

**Test coverage is thorough:**
- Shell metacharacter injection (`;`, `$()`) — skipped with WARN ✅
- Newline entry forgery — newline in directory name doesn't produce extra log lines ✅
- Unit tests for `isSafeImpactedLogPath` — 5 accept cases, 19 reject cases covering every metacharacter class ✅
- Not-configured and blank `scalpel.impactedLog` — no file written ✅

**Minor observation** (not blocking): There's no integration test for the reactor-root-as-`"."` end-to-end path (where the parent project itself is affected and its empty relative path maps to `"."`). The unit tests cover `"."` implicitly through the `"m"` single-char case, and the mapping logic is trivial, but an explicit test would close the gap.

### 📋 PR Metadata

Labels look correct (`bug`, `documentation`, `java`).

Issue #127, opened 2026-08-24
[arch] core's published API surface is accreting test-only members; no binary-compatibility check; standalone usage gaps


paoloantinori:
## Problem

`core` publishes six public types to Maven Central with no japicmp/revapi check, and the surface already carries test-only or dead members: an `analyzeChanges` overload called only from its own test, a `computeBuildSet` overload called only from a test, a `mapToProjects` method and a `TrimResult` constructor with zero callers, telescoping `AffectedModule` constructors used only by tests, and the deprecated unreferenced `REASON_UPSTREAM_DEPENDENCY`.

Four coupling issues in the same area:

- `detectBaseBranch` reads environment variables through the Maven-specific `env.` property prefix with no `System.getenv()` fallback, so a standalone consumer passing `System.getProperties()` gets nothing.
- The reason/category report constants are extension-layer vocabulary living in `core`.
- `Version.version()` hardcodes `core`'s own pom.properties, so `extension3` reports `core`'s version.
- `core` declares `javax.inject` at compile scope while sisu is provided, leaking it onto every consumer's classpath.

## Proposal

Remove or make package-private the dead/test-only members; add japicmp or revapi on `core` bound to `verify`; `System.getenv()` fallback for `detectBaseBranch`; `javax.inject` provided scope; a decision record on where the report vocabulary lives.

gnodet:
Binary-compatibility checks, API surface audits, and test-only member segregation are enterprise-grade governance measures. Scalpel has a small contributor base and isn't a library with downstream consumers who depend on API stability guarantees. The current approach — careful code review — is proportional to the project's scale. Closing as won't-fix.

Issue #82, opened 2026-08-24
[bug] Swallowed IO errors bias the analysis toward under-building (the unsafe direction)


paoloantinori:
## Behavior

Several catch blocks in the analysis path bias the result toward **under-building**, which for a tool whose worst failure mode is a green build on untested code is the unsafe direction. The same files apply the conservative (over-build) posture elsewhere, so this is inconsistent as well as risky.

Sites (checked against current main):

- `PomChangeAnalyzer` `checkFileForPropertyRefs`: catches `IOException` with a "skip unreadable files" comment and no logging. An unreadable filtered resource is treated as NOT containing the changed property, so the module is not marked affected. Two branches earlier, an oversized file returns *affected*: opposite risk postures in the same method.
- `scanDirectoryForPropertyRefs`: skips an unreadable directory with no logging.
- `isBinaryFile`: I/O failure swallowed with no logging.
- `readPomText`: returns null on failure with only a debug log; the caller then silently skips the whole "child references changed property" check.
- `getChangedTransitiveDependencyScope` (participant): a `DependencyResolutionException` is logged at debug with no partial result, and the module is treated as unaffected by a changed managed dependency.

## Expected

Each site should either mark the module affected or at least log at WARN, never silently under-build. `PomChangeAnalyzer` already has the right template in its catch-all that conservatively marks all dependents affected.

A per-build counter of swallowed I/O errors, surfaced in the summary log line and the JSON report, would let CI alert on degraded analysis.

## Suggested fix

I have a branch with a failing-test-first implementation covering six sites (unreadable resource, unreadable directory, resolution failure, etc., all asserting the conservative outcome). Happy to rebase and PR it if the direction is welcome.

paoloantinori:
Adding a deterministic reproduction technique to this one: a Byteman fault-injection rule (test-scope, profile-gated, zero impact on the normal suite) that makes the filtered-resource file unreadable exactly at the `checkFileForPropertyRefs` entry. The real code then throws the real IOException into the real catch, and the outcome is observable without any filesystem-permission setup, so it works the same on any OS and in root containers (where chmod-based tests are vacuous). Result on current main: the module is silently NOT affected, i.e. exactly the under-build described above, now reproducible on demand. I'll include this harness in the PR as an optional `-Pbyteman` profile so the conservative behavior is provable after the fix.

paoloantinori:
[...]

PR #153, opened 2026-08-27
Fix #98: cache scalpel regex patterns and cap match input length

### LINKED ISSUE #98: [security] Regex options matched against attacker-influenced input with no cost bound or pattern cache
## Behavior

`disableOnBranch`, `disableOnBaseBranch` and `forceBuildModules` are user-supplied regexes matched via `String.matches()` inside a `PatternSyntaxException` catch, with no backtracking guard, timeout, input-length cap or compiled-pattern cache.

`forceBuildModules` is the interesting one: it is matched against `project.getArtifactId()`, which a PR author fully controls including length. A maintainer pattern with nested quantifiers plus a crafted artifactId gives exponential backtracking and hangs the Maven JVM. Documented real-world patterns like `.*-it` are linear, so exploitation needs a specific unlucky pattern (whether any real configuration is vulnerable is an open question worth checking). Globs are not fully immune either: Java's glob translation produces chained `.*` segments for multi-wildcard patterns, which is polynomial.

## Proposal

- Cap input length before matching with a documented limit and a WARN when exceeded.
- Compile each pattern once into a cached `Pattern` instead of per call (also a small perf win).
- Either a step-limited `CharSequence` wrapper, or migrate `forceBuildModules` to glob/exact-match semantics.
- Document that these options are user-supplied regexes evaluated against attacker-influenced input.

paoloantinori:
## Problem

As filed in #98: the three regex options (`disableOnBranch`, `disableOnBaseBranch`, `forceBuildModules`) were matched via `String.matches()`, which compiles the pattern on every call - `forceBuildModules` per module per pattern against `project.getArtifactId()`, which a PR author fully controls including length. A maintainer pattern with nested quantifiers plus a crafted artifactId can backtrack exponentially and hang the Maven JVM. Java's glob translation also produces chained `.*` segments (polynomial) for multi-wildcard patterns.

## Change

A new `BoundedRegexMatcher` (one instance per consuming class) now backs both match sites:

- **Compile-once cache**: patterns are compiled into a bounded cache (256 entries; beyond it, degrades silently to per-call compilation, the previous behavior - maintainer-supplied config, so compile cost is not attacker-influenced).
- **Input cap**: branch names and artifactIds longer than 256 characters are skipped with a single WARN naming the config key and the limit, and count as non-matches. The limit is generous against reality: real artifactIds sit well under 100 characters, git refs are conventionally bounded near 255 bytes, and a 256+ character artifactId cannot even pass `mvn install` on mainstream filesystems (filename-component cap of 255 on APFS/ext4/XFS, verified empirically).
- **Invalid/null patterns** are non-matches with one WARN per pattern per key (previously the invalid-pattern WARN fired on every call - log-spam reduction, message text unchanged).
- README documents that these options are user-supplied regexes evaluated against PR-influenced input, with the cap and linear-pattern guidance.

**Design note, flagged for review**: the cap direction is uniform (over-limit = non-match). For the disable options that fails toward "not disabled" (build proceeds, the safe direction per #82's convention). For `forceBuildModules` it means not force-included; the under-build window requires a pre-existing, unchanged module with a 257+ character artifactId that the maintainer force-listed - installable nowhere mainstream and WARNed once. If you weigh "user explicitly asked for this module to always build" as inviolable, a one-line split (force-include on over-limit for forceBuild only) is easy to concede.

**API note**: `BoundedRegexMatcher` is a new public core class (used cross-module by extension3.internal; package-private is impossible cross-artifact). It is exactly the API-surface accretion #127 complains about - happy to relocate to a `core.internal` package as follow-up if you prefer.

## Verification

TDD: RED first (over-limit artifactId force-included and over-limit branch disabling detection, both failing on main; matcher class absent), then GREEN. Full suites green (core 159, extension3 167). Null/empty variants covered for all three option lists; unset maps to the same empty list by parseList.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safer handling for branch and artifact matching patterns.
  * Overly long inputs are skipped and reported with a warning instead of being matched.
  * Invalid patterns no longer interrupt processing and are reported once.
  * Empty matching settings now behave like unset values.
  * Base branch names with remote prefixes are handled correctly.

* **Documentation**
  * Documented matching limits, warning behavior, and recommendations for regex patterns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,6 +391,18 @@ Use `forceBuildModules` with regex patterns matching artifactIds:
 -Dscalpel.forceBuildModules=.*-it,.*-tests
 ```
 
+Patterns use full-match Java regex semantics against the module's `artifactId`. These regexes and
+`disableOnBranch`/`disableOnBaseBranch` are evaluated against input a pull request author
+influences (artifactIds, branch names), so their match cost is bounded:
+
+- patterns are compiled once per build and cached, not compiled per match attempt;
+- inputs longer than **256 characters** are never matched: the attempt is skipped, a WARN is
+  logged once per input, and the input counts as a non-match.
+
+Keep patterns linear: avoid nested quantifiers such as `(a+)+` or `(.*)*`, which can backtrack
+exponentially even against short input. The documented examples above (`.*-it`, `.*-tests`) are
+linear and safe.
+
 For scheduled/cron builds where no changes may be detected, use `buildAllIfNoChanges` to fall back
 to a full build instead of building nothing:
 
@@ -420,7 +432,10 @@ for CI systems where incremental builds should be skipped on main, release, or m
 ```
 
 Branch patterns are Java regular expressions. For `disableOnBaseBranch`, the remote prefix
-(e.g. `origin/`) is stripped before matching.
+(e.g. `origin/`) is stripped before matching. Branch names are pull-request-influenced input,
+so the same cost bounds as `forceBuildModules` apply: patterns are compiled once per build,
+inputs longer than 256 characters are skipped with a WARN (counted as a non-match), and linear
+patterns without nested quantifiers are strongly preferred.
 
 ### Selected Projects (`-pl`) Handling
 
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/BoundedRegexMatcher.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/BoundedRegexMatcher.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) Maveniverse Org.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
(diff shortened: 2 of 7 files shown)


gnodet (approved):
Well-designed security fix for issue #98, bounding regex cost by caching compiled patterns and capping input length. TDD approach with thorough tests, all CI green across platforms.

Two low-severity observations (non-blocking):

1. **Soft bound race** — the cache and warned-set size checks (`cache.size() < MAX_CACHED_PATTERNS`) are not atomic with the subsequent `putIfAbsent`/`add`, so under high concurrency the collections could slightly exceed their stated bounds. Acceptable because these are soft memory bounds, not security-critical invariants — the Javadoc already acknowledges this for the pattern cache.

2. **captureStderr duplication** — the stderr-capture technique for verifying WARN log output is duplicated across three test classes. A shared test utility could reduce duplication if desired in a follow-up.

The design note about the "cap direction" (over-limit = non-match uniformly) is well-reasoned. For `disableOnBranch`/`disableOnBaseBranch`, non-match means build proceeds (safe direction). For `forceBuildModules`, non-match means not force-included, requiring a 257+ char artifactId — practically impossible.

PR #155, opened 2026-08-28
Fix #105: size-cap git blob reads for old-POM comparison

### LINKED ISSUE #105: [security] No size cap on git blob and POM reads
## Problem

`GitChangeDetector.readFileAtCommit` reads a whole blob into a ByteArrayOutputStream and `PomChangeAnalyzer.readPomText` reads a whole pom.xml into a String, both with no size ceiling; `readPomFilesAtCommit` holds every reactor POM in a map simultaneously. The realistic outcome is memory pressure rather than a clean OOM primitive (the merge-base POM comes from the trusted base branch, and Maven has already parsed the PR's own POM), so this deserves a cheap guard rather than a redesign.

Note the recent optimization that narrows `readPomFilesAtCommit` to only changed POMs already removes most of the exposure as a side effect; a cap would close the rest.

## Proposal

- A size ceiling applied before `loader.copyTo` / `Files.readAllBytes`.
- Oversized files skipped with WARN and treated in the conservative (affected) direction.
- Ceiling configurable, reusing `maxResourceFileSize` or a dedicated property.

paoloantinori:
## Problem

As filed in #105: `readPomFilesAtCommit` read whole git blobs into memory with no ceiling (and `readFileAtCommit` the same), with every reactor POM held simultaneously during analysis. The realistic outcome was memory pressure rather than a clean OOM primitive, but a cap is cheap defence. The `scalpel.maxResourceFileSize` property (default 10 MiB) was fully orphaned after the effective-model rework - declared, validated, zero production consumers.

## Change

- Both blob readers in `GitChangeDetector` take the cap, check `ObjectLoader.getSize()` before any copy, and on oversize emit a WARN naming the path, size, cap and the `-Dscalpel.maxResourceFileSize` escape hatch, then skip (`readFileAtCommit` returns null, `readPomFilesAtCommit` omits the entry).
- The omission flows into the existing conservative branch: a changed POM with no base content marks the project and ALL transitive dependents affected (the same branch as a genuinely-new POM). Non-positive caps are rejected up front, mirroring the configuration-layer validation.
- `ScalpelCore` passes `config.getMaxResourceFileSize()`, bringing the property back into service for blob reads.

**Which knob governs which read** (worth stating plainly since the names are close): `scalpel.maxResourceFileSize` (10 MiB, configurable) caps git blob reads for old-POM comparison - this change. The filtered-resource file scan keeps its separate 1 MiB hardcoded limit; PR #129 fixes that side's skip direction. The property's documented meaning changes accordingly (it was dead code before this change).

**Binary-compatibility note**: `readFileAtCommit` is public core API (documented in #149) and its signature gains a parameter - a breaking change for direct users, worth a release-note line.

## Verification

TDD: RED first (the property ignored end-to-end, oversize POM present in old contents), then the fix, then a review pass (the explain evidence no longer misattributes an oversize skip as a "new pom"). Full suites green (core 156, extension3 166). Null-input covered: unset/default 10 MiB, small, non-positive at both config and detector layer. TOCTOU excluded by git's content-addressed immutability (verified against the JGit 7.7.1 jar).

## Summary by CodeRabbit

* **New Features**
  * Added configurable size limits for files read during change analysis.
  * Added detection and actionable warnings for shallow repositories.

* **Bug Fixes**
  * Oversized files are skipped with a warning and conservatively treated as affected.
  * Oversized previous POM files also affect dependent modules.
  * Unreadable files and failed model analysis now conservatively mark affected modules.
  * Files at or below the configured limit continue to be analyzed normally.

* **Documentation**
  * Clarified the maximum resource file size setting and behavior.

### DIFF
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
@@ -153,7 +153,22 @@ public class GitChangeDetector {
         return changedFiles;
     }
 
-    public byte[] readFileAtCommit(Repository repository, ObjectId commitId, String path) throws IOException {
+    /**
+     * Reads the file content at the given path in the commit's tree.
+     * <p>
+     * The blob is read only when its size does not exceed {@code maxFileSize}: an oversized
+     * blob is skipped with a WARN and {@code null} is returned, the same value used for a
+     * path absent at the commit, so callers treat it conservatively (as changed/affected)
+     * instead of loading unbounded content into memory.
+     *
+     * @param maxFileSize maximum blob size in bytes to read (the
+     * {@code scalpel.maxResourceFileSize} cap); must be positive
+     * @return the file content, or null when the path is absent at the commit or its blob
+     * exceeds {@code maxFileSize}
+     */
+    public byte[] readFileAtCommit(Repository repository, ObjectId commitId, String path, long maxFileSize)
+            throws IOException {
+        requirePositiveMaxFileSize(maxFileSize);
         try (RevWalk revWalk = new RevWalk(repository)) {
             RevCommit commit = revWalk.parseCommit(commitId);
             try (TreeWalk treeWalk = TreeWalk.forPath(repository, path, commit.getTree())) {
@@ -161,6 +176,9 @@ public class GitChangeDetector {
                     return null;
                 }
                 ObjectLoader loader = repository.open(treeWalk.getObjectId(0));
+                if (isOverSizeCap(path, loader.getSize(), maxFileSize)) {
+                    return null;
+                }
                 ByteArrayOutputStream out = new ByteArrayOutputStream();
                 loader.copyTo(out);
                 return out.toByteArray();
@@ -169,6 +187,26 @@ public class GitChangeDetector {
             warnIfShallow(repository, commitId.getName(), path);
             throw e;
         }
+    }
+
+    private static void requirePositiveMaxFileSize(long maxFileSize) {
+        if (maxFileSize <= 0) {
+            throw new IllegalArgumentException("maxFileSize must be positive, got " + maxFileSize);
+        }
+    }
+
+    private boolean isOverSizeCap(String path, long blobSize, long maxFileSize) {
+        if (blobSize > maxFileSize) {
+            logger.warn(
+                    "Skipping blob {}: size {} bytes exceeds scalpel.maxResourceFileSize ({} bytes);"
+                            + " treating the change conservatively as affected."
+                            + " Raise -Dscalpel.maxResourceFileSize to read this blob",
+                    path,
+                    blobSize,
+                    maxFileSize);
+            return true;
+        }
+        return false;
     }
 
     /**
@@ -274,9 +312,18 @@ public class GitChangeDetector {
      * Reads multiple files at the given commit in a single TreeWalk pass.
      * Uses a PathFilterGroup so the TreeWalk only visits matching entries,
      * and shares a single ObjectReader/RevWalk across all reads.
+     * <p>
+     * Blobs larger than {@code maxFileSize} are skipped with a WARN and omitted from the
+     * result, the same treatment as a path absent at the commit: POM-change analysis then
+     * marks the module and its dependents as affected rather than loading unbounded
+     * content into memory.
+     *
+     * @param maxFileSize maximum blob size in bytes to read (the
+     * {@code scalpel.maxResourceFileSize} cap); must be positive
      */
-    public Map<String, byte[]> readPomFilesAtCommit(Repository repository, ObjectId commitId, Set<String> paths)
-            throws IOException {
+    public Map<String, byte[]> readPomFilesAtCommit(
+            Repository repository, ObjectId commitId, Set<String> paths, long maxFileSize) throws IOException {
+        requirePositiveMaxFileSize(maxFileSize);
         if (paths.isEmpty()) {
             return Map.of();
         }
@@ -295,6 +342,9 @@ public class GitChangeDetector {
                             || treeWalk.getFileMode(0) == FileMode.EXECUTABLE_FILE) {
                         String path = treeWalk.getPathString();
                         ObjectLoader loader = reader.open(treeWalk.getObjectId(0));
+                        if (isOverSizeCap(path, loader.getSize(), maxFileSize)) {
+                            continue;
+                        }
                         ByteArrayOutputStream out = new ByteArrayOutputStream();
                         loader.copyTo(out);
                         result.put(path, out.toByteArray());
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
@@ -166,6 +166,14 @@ public class GitChangeDetector {
      * @return the file content, or null when the path is absent at the commit or its blob
      * exceeds {@code maxFileSize}
      */
+    /**
+     * Backward-compatible overload delegating to {@link #readFileAtCommit(Repository, ObjectId, String, long)}
+     * with the documented default cap ({@code scalpel.maxResourceFileSize}, 10 MiB).
+     */
+    public byte[] readFileAtCommit(Repository repository, ObjectId commitId, String path) throws IOException {
+        return readFileAtCommit(repository, commitId, path, ScalpelConfiguration.DEFAULT_MAX_RESOURCE_FILE_SIZE);
+    }
+
     public byte[] readFileAtCommit(Repository repository, ObjectId commitId, String path, long maxFileSize)
             throws IOException {
         requirePositiveMaxFileSize(maxFileSize);
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
@@ -154,6 +154,14 @@ public class GitChangeDetector {
     }
 
     /**
+     * Backward-compatible overload delegating to {@link #readFileAtCommit(Repository, ObjectId, String, long)}
+     * with the documented default cap ({@code scalpel.maxResourceFileSize}, 10 MiB).
+     */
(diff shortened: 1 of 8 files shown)


core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
```diff
@@ -121,21 +121,59 @@ public Set<String> getChangedFiles(Repository repository, ObjectId base, ObjectI
         return changedFiles;
     }
 
-    public byte[] readFileAtCommit(Repository repository, ObjectId commitId, String path) throws IOException {
+    /**
+     * Reads the file content at the given path in the commit's tree.
```
gnodet, line 165: **Design suggestion (medium):** `readFileAtCommit` gains a mandatory `maxFileSize` parameter, removing the old 3-arg public API signature. The PR description correctly flags this as a breaking change. Since this method is documented public API (see #149) and the project has published releases, consider preserving a backward-compatible overload:

```java
public byte[] readFileAtCommit(Repository repository, ObjectId commitId, String path) throws IOException {
    return readFileAtCommit(repository, commitId, path, ScalpelConfiguration.DEFAULT_MAX_RESOURCE_FILE_SIZE);
}
```

Same applies to `readPomFilesAtCommit`. This avoids a binary-incompatible break for any external consumers while still exposing the new cap parameter.

core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelCoreTest.java
```diff
@@ -273,6 +275,98 @@ void detectChanges_readOldPomContent() throws Exception {
...
+                    new String(result.getOldPomContents().get("pom.xml"), StandardCharsets.UTF_8).contains("1.0"),
+                    "old POM content must be the base-branch version");
+        }
+    }
+
+    /**
+     * Runs the callable with System.err captured (slf4j-simple writes to stderr)
+     * and returns everything that was written during its execution.
+     */
+    private String captureStderr(Runnable action) {
```
gnodet: **Minor (low):** The `captureStderr` helper is duplicated — it already exists identically in `GitChangeDetectorTest.java` (line 737). Consider extracting it to a shared test utility class to reduce duplication.

gnodet (commented):
Well-structured TDD fix that caps git blob reads at `scalpel.maxResourceFileSize`, closing a real unbounded-memory gap (#105). The implementation is correct and conservative (oversized blobs treated as affected). Two design suggestions below.

**Highlights:**
- The TDD commit sequence (RED test → fix → review cleanup) is exemplary and makes the change easy to verify.
- The conservative direction (oversized blob → affected) is the right design choice — it may over-build but never under-build.
- The `isOverSizeCap` WARN message is well-crafted: it names the path, blob size, configured cap, and the `-Dscalpel.maxResourceFileSize` escape hatch, giving operators actionable information.
- The `PomChangeAnalyzer` wording update from "new pom" to "no base content (new or oversize-skipped)" accurately covers both cases.
- The `requirePositiveMaxFileSize` validation at the detector layer provides defense-in-depth alongside the `ScalpelConfiguration` validation.
- Boundary behavior (blob size == cap is allowed, only strictly greater is skipped) is correctly tested.

### 📋 PR Metadata

Labels look correct (`bug` for #105, `java` auto-applied).

core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
paoloantinori: Added in the follow-up commit: a 3-arg overload delegating with ScalpelConfiguration.DEFAULT_MAX_RESOURCE_FILE_SIZE, plus a test asserting the backward-compatible form reads a file under the default cap.

core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelCoreTest.java
paoloantinori: Extracted to a shared TestOutputCapture utility; both test classes delegate to it.

gnodet (dismissed):
**Re-review after new commits.** Both previous findings are addressed. CI is fully green.

**Previous findings resolved:**
- ✅ **(medium) Backward-compatible overload**: The 3-arg `readFileAtCommit` signature is preserved as an overload delegating with `DEFAULT_MAX_RESOURCE_FILE_SIZE`, with a test confirming it works (`readFileAtCommit_threeArgOverload_usesDefaultCap`). `readPomFilesAtCommit` still has the new signature only, but it has only one production caller (`ScalpelCore`) and appears to be treated as internal API — acceptable.
- ✅ **(low) Duplicated captureStderr**: Extracted to shared `TestOutputCapture` utility; both `GitChangeDetectorTest` and `ScalpelCoreTest` delegate to it. Clean implementation with proper license header and package-private visibility.

**One minor note:**
- The detailed Javadoc block (with `@param maxFileSize`, `@return`) at line ~156 is positioned above the 3-arg backward-compatible overload, but it describes the 4-arg method. In Java, only the last doc comment before a declaration applies, so the 4-arg primary method at line 177 ends up with no Javadoc. Consider moving that block directly above the 4-arg method.

### 📋 PR Metadata

Labels look correct (`bug` for #105, `java` auto-applied).

gnodet (commented):
Sound implementation of blob-size capping for git reads with thorough test coverage. Two concrete issues worth addressing:

1. **Orphaned javadoc** — Two consecutive javadoc blocks appear before the 3-arg `readFileAtCommit` overload. The first block (documenting `@param maxFileSize` for the 4-arg overload) is orphaned because Java attaches javadoc only to the immediately following declaration. The 4-arg method consequently has no javadoc.

2. **Binary compatibility** — `readPomFilesAtCommit` changes from 3 args to 4 args without a backward-compatible overload, unlike `readFileAtCommit` which correctly provides one. For consistency, either add a 3-arg overload or explicitly document the break.

The core implementation is well-structured: `requirePositiveMaxFileSize` and `isOverSizeCap` are cleanly factored, the warning message is actionable, and the conservative fallback (null/omit) integrates naturally. The `TestOutputCapture` extraction is a clean deduplication.

### 📋 PR Metadata

|--------|---------|-----------|

core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
```diff
@@ -153,14 +153,40 @@ public Set<String> getChangedFiles(Repository repository, ObjectId base, ObjectI
         return changedFiles;
     }
 
+    /**
```
gnodet, line 164: Two consecutive javadoc blocks appear before the 3-arg `readFileAtCommit` overload. The first block (containing `@param maxFileSize` and `@return`) is intended for the 4-arg overload but is orphaned — Java attaches only the immediately preceding javadoc to a declaration. The 4-arg `readFileAtCommit(Repository, ObjectId, String, long)` consequently has no javadoc at all.

**Fix:** Move the first javadoc block below the 3-arg method so it sits directly above the 4-arg method.

core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
```diff
@@ -274,9 +320,18 @@ public void fetchBranch(Repository repository, String baseBranch) throws IOExcep
...
+     * result, the same treatment as a path absent at the commit: POM-change analysis then
+     * marks the module and its dependents as affected rather than loading unbounded
+     * content into memory.
+     *
+     * @param maxFileSize maximum blob size in bytes to read (the
+     * {@code scalpel.maxResourceFileSize} cap); must be positive
      */
-    public Map<String, byte[]> readPomFilesAtCommit(Repository repository, ObjectId commitId, Set<String> paths)
-            throws IOException {
+    public Map<String, byte[]> readPomFilesAtCommit(
```
gnodet, line 342: `readPomFilesAtCommit` is public API (documented via #149). Its signature changed from 3 args to 4 args without a backward-compatible overload, breaking binary compatibility. The PR correctly provides a 3-arg overload for `readFileAtCommit` to maintain compatibility, but `readPomFilesAtCommit` does not receive the same treatment.

Consider adding a backward-compatible 3-arg overload that delegates with `DEFAULT_MAX_RESOURCE_FILE_SIZE`, or explicitly document in the PR description that both methods break compatibility.

core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
gnodet: Fixed in 3c0c325. Moved the detailed javadoc block (with `@param maxFileSize`) directly above the 4-arg `readFileAtCommit` method. The 3-arg overload keeps its own brief javadoc.

core/src/main/java/eu/maveniverse/maven/scalpel/core/GitChangeDetector.java
gnodet: Fixed in 3c0c325. Added a 3-arg `readPomFilesAtCommit` overload delegating with `DEFAULT_MAX_RESOURCE_FILE_SIZE`, matching the `readFileAtCommit` pattern.

gnodet (approved):
**Re-review after commit 3c0c325.** All previous findings are resolved:

- ✅ **Orphaned javadoc**: The detailed javadoc block (with `@param maxFileSize`, `@return`) is now correctly positioned directly above the 4-arg `readFileAtCommit` method. The 3-arg overload has its own brief javadoc. Clean.
- ✅ **`readPomFilesAtCommit` binary compatibility**: 3-arg overload added, delegating with `DEFAULT_MAX_RESOURCE_FILE_SIZE`, matching the `readFileAtCommit` pattern.
- ✅ **`readFileAtCommit` backward-compatible overload**: Already addressed in prior commit.
- ✅ **`captureStderr` deduplication**: `TestOutputCapture` shared utility, both test classes delegate.

The fix commit is minimal and correct — no new issues introduced. The overall PR is solid: size-cap logic is clean, validation is defense-in-depth, the conservative fallback direction (oversized → affected) is correct, and test coverage is thorough (boundary, oversize, non-positive cap, 3-arg overload backward compat, end-to-end through ScalpelCore).

### 📋 PR Metadata

Labels look correct (`bug`, `java`).

PR #164, opened 2026-09-04
Fix #92: shadow mode measuring savings and false negatives from one full build

### LINKED ISSUE #92: [enh] Shadow mode: measure savings and false negatives from a single full build
## Problem

A team can only discover how much Scalpel would save on THEIR reactor topology by building a bespoke harness. Two of five known adopters independently built one in bash: Apache Camel computes `scalpel_would_test`/`scalpel_would_skip` with jq and renders a PR comment; Apicurio built a 7-shard shadow matrix plus classification and accuracy scripts. Neither harness is reusable by the other; both re-derive the same primitives.

The cost of not having it built in: Apicurio spent ~100 runner-hours/week for ~18 days to learn its hub-and-spoke topology yields 0% beneficiary PRs, then retired Scalpel from gating. (That same evaluation produced the strongest correctness datum Scalpel has: zero false negatives across 121 control-versus-shadow comparisons.)

## Proposal

`-Dscalpel.mode=shadow`: behaves exactly like `report` (full build, reactor untouched, all tests run) and additionally:

1. Computes the trim/skip-tests decision it WOULD have made and writes it with the full metrics block.
2. Registers an `ExecutionListener` recording per-module wall-clock for the full build actually running.
3. Joins the two: modules it would have skipped x measured duration = `estimatedSecondsSaved`, from a single full build with no control group.
4. Records `wouldHaveSkippedButFailed`: modules it would have skipped that actually failed - the false-negative counter, computed for free on every shadow run.
5. Appends one JSONL line per run to `target/scalpel-shadow-history.jsonl` for trend analysis.

Would pair naturally with the timing instrumentation idea. Happy to help build it.

paoloantinori:
Adds `-Dscalpel.mode=shadow`: behaves exactly like `report` (full build, reactor untouched, all tests run, the standard JSON report is written) and additionally measures the build, so a team learns what Scalpel would save on its topology from a single full build, with no control group and no extra runner-hours (the two bespoke bash harnesses Camel and Apicurio built, and the ~100 runner-hours/week Apicurio paid, are the motivation in #92).

How it works:

1. Computes the trim decision it WOULD have made, via the same `ReactorTrimmer.computeBuildSet` call trim mode runs on the same inputs, so shadow and trim decisions agree by construction (the report branch's existing trimResult is reused when the input sets are identical)
2. Wraps the session's existing `ExecutionListener` in a `ShadowBuildMonitor` recording per-module wall-clock and failures; every event is forwarded to the wrapped listener unchanged
3. At session end joins the two into `target/scalpel-shadow.json`: `estimatedSecondsSaved` (skipped modules x measured duration) and `wouldHaveSkippedButFailed` (modules it would have skipped that failed, the false-negative counter, populated even on failed builds)
4. Appends one JSONL line per measured run to `target/scalpel-shadow-history.jsonl`

TDD: the RED commit (config acceptance, monitor unit tests, three ITs) fails on unmodified main; then the implementation. Gates run before pushing: a four-angle simplify pass (applied: injected module keying so the decision and the monitor share one key algorithm, trimResult reuse, delegate normalization, single join computation; skipped with reasons: sharing the JSON escaping helpers would grow the published core API they are private to) and a five-angle review (applied: bail-outs now overwrite the shadow document with a status-only document so a previous run's measurement can never survive as current, mirroring the report's #89 semantics; `changedFilesCount` numeric in both artifacts; root aggregator keyed as `."; parity claim scoped; field table with types documented; drift-guard test pins the emitted field set).

* shadow replaces the session's `ExecutionListener` (wrapped, all events forwarded to any pre-existing listener, including null)
* shadow also writes the standard JSON report and the status-only report on bail-outs, exactly like report mode
* when a shadow run bails out before measuring (no base branch, not a git repo, disable/full-build triggers, fail-safe), `target/scalpel-shadow.json` is overwritten with a status document (`status`/`reason`); the history file is appended only by measured runs, so a gap means not-measured
* `isPassiveMode()` joins the published `ScalpelConfiguration` API (it drives 11 report-artifact gates in the participant; the alternative, inlining `isModeReport() || isModeShadow()` at each, seemed worse)

**Deliberately out of scope** (happy to follow up if wanted): configurable shadow output paths (`scalpel.shadowFile`/`shadowHistoryFile`), a versioned JSON schema for the shadow document like the report has, and history lines for skipped runs. The issue text fixes the artifact paths and does not ask for a schema, so those felt like maintainer calls.

Three new ITs: shadow-mode (reactor untouched, report also written, decision parity, one history line), shadow-failure (planted failing test in a would-be-skipped module; shadow document written on the failed build with the false negative flagged), shadow-status (stale document overwritten on a bail-out). Full battery green: core 197, extension3 188 unit tests, all 32 extension3 ITs.

## Summary by CodeRabbit

* **New Features**
  * Added Shadow mode, which runs the full build while recording trim decisions, module durations, estimated time savings, and potential false negatives.
  * Added JSON and JSONL shadow reports, including status documents when runs cannot be measured.
  * Shadow mode preserves standard build reports and does not trim reactor modules.
  * Shadow decisions honor project-selection and include-path settings.

* **Documentation**
  * Documented Shadow mode configuration, output formats, adoption guidance, and bail-out behavior.


paoloantinori:
Applied the outside-diff finding in 4963ba3: the `-pl` disable exit and the fail-safe `ScalpelException` catch now overwrite `target/scalpel-shadow.json` with a status document (skipped / failed), closing the last two unmeasured exits that could leave a previous run's measurement in place. Two new ITs pin it (`shadow-pl`, planted stale document; `shadow-failsafe`, unresolvable base branch, kept as a regression pin for the status-report routing). All 34 ITs green locally.

One adjacent observation outside this PR's scope, for you to call: the same fail-safe `ScalpelException` catch leaves the JSON **report** stale in report mode too (pre-existing behaviour; the generic `Exception` catch beside it does write a failed-status report). Happy to align it with `writeFailedStatusReport` in a one-liner follow-up if you want the #89 invariant complete there as well.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -485,6 +495,61 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
...
+                    Set<String> wouldHaveBuilt = new LinkedHashSet<>();
+                    java.util.function.Function<MavenProject, String> moduleKey = project -> {
+                        String path = relativePath(reactorRoot, project);
+                        // The root aggregator relativizes to the empty string; report it as
+                        // "." like the impacted log does (#84) so every module has a name.
+                        return path.isEmpty() ? "." : path;
+                    };
+                    for (MavenProject project : decision.getBuildSet()) {
+                        wouldHaveBuilt.add(moduleKey.apply(project));
+                    }
```
gnodet, line 548: ⚠️ **`includePaths` parity gap** — trim mode applies a post-`computeBuildSet` filter (line ~593) that removes downstream modules outside `includePaths` scope from the build set. Shadow populates `wouldHaveBuilt` directly from `decision.getBuildSet()` without this filter.

The comment above says the two "agree by construction" — that's true for the `computeBuildSet` call itself, but trim mode's actual build set is a *subset* of `getBuildSet()` when `includePaths` is configured. Shadow's `wouldHaveSkipped` is therefore smaller than what trim actually skips, and `estimatedSecondsSaved` undercounts.

This is conservatively wrong (underestimates savings rather than overclaiming), but it breaks the parity promise and produces confusing results when a team configures `includePaths` and sees shadow savings ≠ actual trim savings.

Fix: mirror the `includePaths` post-filter here before populating `wouldHaveBuilt`, same predicate as the trim path.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -1532,12 +1603,37 @@ private void writeStatusReport(ScalpelConfiguration config, Path reactorRoot, St
                     status,
                     reason,
                     config.getReportFile());
+            if (config.isModeShadow()) {
+                writeShadowStatus(reactorRoot, status, reason);
```
gnodet: 💡 **`writeShadowStatus` coupled to report write success** — this call sits after `report.writeToFile(...)` (line 1600) inside the same `try` block. If `writeToFile` throws (e.g., custom `scalpel.reportFile` path is unwritable), the shadow status write is skipped and a stale shadow document survives — exactly the situation the #89 invariant is designed to prevent.

Since `writeShadowStatus` has its own independent `try/catch`, it can safely run regardless of the report write outcome.

```suggestion
            if (config.isModeShadow()) {
                writeShadowStatus(reactorRoot, status, reason);
            }
            report.writeToFile(reactorRoot, config.getReportFile());
```

Alternatively, move it to a `finally` block. The same pattern applies in `writeFullBuildReport` (line ~1652).

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -1550,6 +1646,11 @@ private void writeFullBuildReport(
         try {
             report.writeToFile(reactorRoot, config.getReportFile());
             logger.info("Scalpel: Report written to {}", config.getReportFile());
+            if (config.isModeShadow()) {
+                // A full build was triggered, so the would-be decision is "build all":
+                // record that instead of leaving a previous run's measurement in place.
+                writeShadowStatus(reactorRoot, "skipped", "full build triggered by " + triggerFile);
+            }
```
gnodet: 💡 Same coupling issue as in `writeStatusReport` — `writeShadowStatus` runs only after `report.writeToFile` succeeds. If the report write fails, the shadow document is not overwritten.

```suggestion
            if (config.isModeShadow()) {
                // A full build was triggered, so the would-be decision is "build all":
                // record that instead of leaving a previous run's measurement in place.
                writeShadowStatus(reactorRoot, "skipped", "full build triggered by " + triggerFile);
            }
```

Move this before `report.writeToFile`, or into a `finally`.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -308,7 +318,7 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
 
             if (directlyAffected.isEmpty() && oldEffectiveModels.isEmpty()) {
                 logger.info("Scalpel: No modules affected by changes");
-                if (config.isModeReport()) {
+                if (config.isPassiveMode()) {
```
gnodet, line 321: 🔵 **Missing shadow status overwrite on `writeReport` early returns** — this `isPassiveMode()` branch calls `writeReport()` and returns without installing `ShadowBuildMonitor` and without overwriting a stale shadow document. A previous shadow run's measurement survives this exit. The same gap exists at lines ~385 and ~430 (the other "no modules affected" / "no modules match `includePaths`" early returns).

The `writeStatusReport`-based exits all got `writeShadowStatus` coverage (correctly), but these `writeReport`-based exits didn't. For #89 consistency, these should also call `writeShadowStatus(reactorRoot, "skipped", ...)` in shadow mode.

Low severity — this only matters in local dev with a surviving `target/` directory from a previous shadow run that did measure savings. In CI with clean workspaces, it's a no-op.

gnodet (commented):
**Shadow mode (#92) — solid feature, three gaps to close**

Well-structured PR with thorough IT coverage (5 ITs covering success, failure, status, -pl, and fail-safe paths). The `ShadowBuildMonitor` class is clean, concurrency-safe, and correctly non-invasive. The #89 stale-doc invariant is systematically applied to most bail-out paths.

Three issues to address before merge:

1. **`includePaths` parity gap** — the shadow decision doesn't mirror the post-`computeBuildSet` filtering that trim mode applies when `includePaths` is configured, so the "by construction" parity claim doesn't hold for that configuration. The savings estimate is conservative (undercount), which is safe but misleading.
2. **`writeShadowStatus` coupled to report write** — in `writeStatusReport` and `writeFullBuildReport`, the shadow status write runs inside the same try block as the report write. A report write failure prevents the shadow status from being written, violating the independence the #89 invariant requires.
3. **`writeReport` early returns skip shadow status** — several bail-out paths (no affected modules, no modules match `includePaths`) use `writeReport()` which doesn't overwrite the shadow document. A previous run's measurement survives these exits.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
paoloantinori: Applied in 5b1eed4: the shadow decision now mirrors trim mode's post-computeBuildSet filter, dropping downstream modules outside the includePaths scope from wouldHaveBuilt. The parity promise now holds for that configuration too, and the savings estimate stops undercounting. New shadow-includepaths IT (cloned from the include-paths fixture, same change classifying module-a affected with module-b downstream and out of scope) pins module-b in wouldHaveSkipped; it was demonstrated failing on the previous commit.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
paoloantinori: Applied as suggested in 5b1eed4: in both writeStatusReport and writeFullBuildReport the shadow overwrite now runs first and independently of the report write, so a failed report write cannot leave a previous run's measurement in place.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
paoloantinori: Applied in 5b1eed4, same first-and-independent ordering as writeStatusReport.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
paoloantinori: Applied in 5b1eed4: all three writeReport-based exits (no modules affected x2, no modules match includePaths) now overwrite the shadow document with a skipped status. The new shadow-unaffected IT plants a stale document and changes a repo-root file (owned by no module since #66), which reaches the no-affected-modules exit; it was demonstrated failing on the previous commit. All 36 ITs green.

gnodet (approved):
**Re-review: all three findings from the previous review are addressed.** ✅

1. **`includePaths` parity gap** — **Addressed** in `5b1eed4`. The shadow decision now applies the same `removeIf` predicate as trim mode's post-`computeBuildSet` filter (line 542 mirrors line 618), dropping downstream modules outside `includePaths` scope from `wouldHaveBuilt`. New `shadow-includepaths` IT pins module-b in `wouldHaveSkipped` — correct.

2. **`writeShadowStatus` coupled to report write** — **Addressed** in `5b1eed4`. In both `writeStatusReport` and `writeFullBuildReport`, `writeShadowStatus` now runs *first* and independently of `report.writeToFile()`. Since `writeShadowStatus` has its own `try/catch`, a report write failure can no longer leave a stale shadow document.

3. **`writeReport` early returns skip shadow status** — **Addressed** in `5b1eed4`. All three `writeReport`-based exits (no affected modules ×2, no modules match `includePaths`) now call `writeShadowStatus(reactorRoot, "skipped", ...)` before `writeReport()` in shadow mode. New `shadow-unaffected` IT plants a stale document and verifies the overwrite.

The new ITs are well-designed: `shadow-includepaths` clones the existing `include-paths` fixture topology and verifies the filtered decision set, `shadow-unaffected` exercises the root-level-file-change scenario (no module ownership since #66). Both properly assert the negative (stale doc gone, history not appended for unmeasured runs).

No new issues found in this commit. The CodeRabbit latest review also found no actionable comments. Good to merge.

PR #165, opened 2026-09-05
Fix #97: contain the report and impacted-log output paths inside the reactor root

### LINKED ISSUE #97: [security] reportFile and impactedLog paths are resolved without containment check
## Behavior

Both output paths call `reactorRoot.resolve(userSuppliedPath)` with no containment check. `Path.resolve(String)` returns the argument unchanged when absolute and does not normalize `../` traversal, and `Files.createDirectories` creates intervening directories anywhere. Since `.mvn/maven.config` in a PR can set any `scalpel.*` property, a PR author controls the destination of a partly-attacker-controlled JSON write and a fully-attacker-controlled line-based log write (near-arbitrary file write, e.g. against a shell startup file).

The write happens in `afterProjectsRead`, before any plugin from the PR runs, so an operator doing a report-only pre-flight pass may believe nothing from the PR has executed yet.

Severity is moderate rather than high only because anyone who can add `.mvn/maven.config` to a build that executes PR content already has near-RCE through Maven itself. This is cheap defence in depth.

Worth preserving and documenting explicitly: `<properties>` in a PR's `pom.xml` do NOT feed Scalpel configuration (only session properties are read) - a good boundary.

## Proposal

- Normalize both paths and reject unless they resolve inside the reactor root; reject absolute paths with a clear message naming the property.
- README note that `mvn validate` on untrusted PR content still executes PR-controlled plugins, so report-only is not a sandbox.

paoloantinori:
Both output paths were resolved with a bare `reactorRoot.resolve(userPath)`: `Path.resolve` returns absolute arguments unchanged and does not normalize parent traversal, so a PR could aim the report write and the fully attacker-controlled impacted-log write outside the project through `.mvn/maven.config` (the issue's verified repro). The write happens in `afterProjectsRead`, before any plugin from the PR runs, which is exactly the report-only pre-flight scenario the issue describes.

`ScalpelReport` gains a public `resolveContained(reactorRoot, configuredPath, propertyName)`: empty, absolute and escaping paths are rejected with an `IOException` naming the property. `writeToFile` routes every report write through it (status and full-build reports included), and the impacted log validates before writing, routing the rejection through the existing fail-safe-aware `handleWriteFailure`. The README carries the requested note that report mode is not a sandbox, plus the confinement rule for both properties in the configuration table.

TDD: the RED commit demonstrates absolute and traversal values for both properties written outside the reactor root with no exception on unmodified main.

Gates run before pushing (simplify four angles, review: bypass scan, null-input audit, history): no bypass vector survives (Windows separators, drive-relative and UNC forms, NUL bytes, URI forms, paths collapsing onto the root; the shadow outputs use fixed constants and the status writers inherit containment through writeToFile); rejection composes with #151's line-level allowlist (destination rejected before any directory is created) and honours the #137/#89 fail-safe semantics at every write site; the explicitly-empty `scalpel.reportFile` case (reachable, since the default does not apply to explicit empties) is now pinned by a test.

* a rejected path means no write at all, so a stale file at the rejected out-of-root location is not overwritten (we refuse to touch it; the in-root #89 overwrite semantics are unchanged)
* an explicitly empty `scalpel.reportFile=` now fails the report write with a clear property-naming message (WARN, build continues under the default failSafe) instead of an obscure directory-write error
* the check is lexical by design and documented as such in the javadoc and README: a pre-existing symlink inside the root pointing elsewhere is not resolved (the issue frames the fix as defence in depth, and `toRealPath` would fail on the not-yet-existing report target)

Full battery green: core 200, extension3 190 unit tests, all 36 integration tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Report and impacted-log paths are prevented from escaping the project directory, including through absolute paths, empty values, or parent-directory traversal.
  * Invalid path configurations now produce warnings or build failures according to fail-safe settings.
  * Relative paths containing `..` remain supported when they normalize within the project.

* **Documentation**
  * Clarified output-path restrictions for reports and impacted logs.
  * Documented handling of unsupported module paths.

### DIFF
--- a/core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelReportTest.java
+++ b/core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelReportTest.java
@@ -197,14 +197,20 @@ class ScalpelReportTest {
     @Test
     void writeToFile_rejectsAbsoluteReportFile() throws IOException {
         ScalpelReport report = minimalReport();
-        Path outside = Files.createTempDirectory("scalpel-outside").resolve("pwned.json");
+        // Created outside JUnit's @TempDir lifecycle, so cleaned up explicitly.
+        Path outsideDir = Files.createTempDirectory("scalpel-outside");
+        try {
+            Path outside = outsideDir.resolve("pwned.json");
 
-        IOException e = assertThrows(IOException.class, () -> report.writeToFile(tempDir, outside.toString()));
+            IOException e = assertThrows(IOException.class, () -> report.writeToFile(tempDir, outside.toString()));
 
-        assertTrue(
-                e.getMessage().contains("scalpel.reportFile"),
-                "The rejection must name the property, got: " + e.getMessage());
-        assertFalse(Files.exists(outside), "Nothing may be written outside the reactor root");
+            assertTrue(
+                    e.getMessage().contains("scalpel.reportFile"),
+                    "The rejection must name the property, got: " + e.getMessage());
+            assertFalse(Files.exists(outside), "Nothing may be written outside the reactor root");
+        } finally {
+            Files.deleteIfExists(outsideDir);
+        }
     }
 
     @Test
--- a/core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelReportTest.java
+++ b/core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelReportTest.java
@@ -195,6 +195,40 @@ class ScalpelReportTest {
     }
 
     @Test
+    void writeToFile_rejectsAbsoluteReportFile() throws IOException {
+        ScalpelReport report = minimalReport();
+        Path outside = Files.createTempDirectory("scalpel-outside").resolve("pwned.json");
+
+        IOException e = assertThrows(IOException.class, () -> report.writeToFile(tempDir, outside.toString()));
+
+        assertTrue(
+                e.getMessage().contains("scalpel.reportFile"),
+                "The rejection must name the property, got: " + e.getMessage());
+        assertFalse(Files.exists(outside), "Nothing may be written outside the reactor root");
+    }
+
+    @Test
+    void writeToFile_rejectsParentTraversalReportFile() throws IOException {
+        ScalpelReport report = minimalReport();
+
+        IOException e = assertThrows(IOException.class, () -> report.writeToFile(tempDir, "../pwned.json"));
+
+        assertTrue(
+                e.getMessage().contains("scalpel.reportFile"),
+                "The rejection must name the property, got: " + e.getMessage());
+        assertFalse(
+                Files.exists(tempDir.getParent().resolve("pwned.json")),
+                "Nothing may be written outside the reactor root");
+    }
+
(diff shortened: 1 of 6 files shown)


core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelReportTest.java
```diff
@@ -194,6 +194,58 @@ void writeToFile_createsFileWithCorrectContent() throws IOException {
         assertEquals(report.toJson(), content);
     }
 
+    @Test
+    void writeToFile_rejectsAbsoluteReportFile() throws IOException {
+        ScalpelReport report = minimalReport();
+        Path outside = Files.createTempDirectory("scalpel-outside").resolve("pwned.json");
```
gnodet: 💡 `Files.createTempDirectory("scalpel-outside")` creates a directory in the system temp area, outside the `@TempDir`-managed lifecycle. JUnit won't clean it up after the test. Since the write is rejected the directory stays empty, but it still leaks on every test run.

Consider creating it as a subdirectory of `tempDir`'s parent (which JUnit also cleans) or adding explicit cleanup:

```suggestion
        Path outsideDir = Files.createTempDirectory("scalpel-outside");
        try {
            Path outside = outsideDir.resolve("pwned.json");
```

with a `finally { Files.deleteIfExists(outsideDir); }` block after the assertions. Same applies to `impactedLog_absolutePathIsRejectedNotWritten` in `ScalpelLifecycleParticipantTest`.

gnodet (commented):
Security-critical path containment logic is solid. `resolveContained` correctly handles null, empty, absolute, traversal, and invalid paths. All write sites are covered (shadow writes use hardcoded constants). Fail-safe integration in the extension is correct. Tests cover the security-relevant attack vectors.

One minor test hygiene issue below.

core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelReportTest.java
paoloantinori: Applied in 678b121: both absolute-path tests now delete the temp directory in a finally, as suggested. The CodeRabbit doc-precision finding is applied in the same commit: README and the property table now state that a path is rejected when its normalization escapes the reactor root (`target/../report.json` stays contained and is allowed) rather than blanket-rejecting `..`.

gnodet (approved):
Re-review of 678b121 — both findings from the previous review are addressed:

1. **Temp directory cleanup** (my finding) — Both `writeToFile_rejectsAbsoluteReportFile` and `impactedLog_absolutePathIsRejectedNotWritten` now wrap the `Files.createTempDirectory` call in try/finally with `Files.deleteIfExists(outsideDir)`. No more leaked temp directories.

2. **Doc precision** (CodeRabbit finding) — README and CONFIGURATION.md now correctly state that paths are rejected when their normalization escapes the reactor root, not when they contain `..` segments in general. `target/../report.json` stays contained and is allowed, which matches `resolveContained`'s actual behavior.

No new issues in this commit. The containment logic in `resolveContained` is unchanged and solid.

PR #167, opened 2026-09-07
Fix #112: optimize filtered-resource property scanning

### LINKED ISSUE #112: [perf] Child POM text re-read and re-scanned per changed property in the parent-POM path
## Problem

For each dependent of a changed parent POM, `readPomText` does a full `Files.readAllBytes` + UTF-8 decode of the child's pom.xml with no memoization; then the loop calls `contains` with a freshly allocated placeholder needle once per changed property, rescanning the whole text each time. A version-bump PR changing 200 properties in a root POM with 500 children = 500 file reads plus on the order of 200 x 500 full-text scans.

Two adjacent micro-issues: an `Xpp3Dom.toString()` serializing the entire plugin configuration DOM inside the per-property loop, and a dependents count computed even when its only consumer is a debug-guarded log.

## Proposal

- Per-invocation POM-text cache keyed by project.
- Replace the N substring searches with one linear scan extracting every property placeholder token into a set, then a single disjointness check: O(P x len) becomes O(len + P).
- Hoist the `Xpp3Dom.toString()`; move the debug-only count inside `isDebugEnabled`.

Would pair naturally with the timing/operation-count instrumentation so the improvement is provable.

gnodet:
## Summary

Fixes #112: `[perf] Child POM text re-read and re-scanned per changed property in the parent-POM path`

### Changes

**Single-pass linear scan for property placeholders**

Replaced the O(P × len) per-property `String.contains()` loop in `checkFileForPropertyRefs` with a single-pass `indexOf`-based scan that extracts all `${...}` placeholder tokens and checks set intersection with changed property names. The new complexity is O(len + P) where `len` is the file content length and `P` is the number of changed properties.

The `containsAnyPropertyRef()` method is package-private (`static`) for direct unit testing.

**Debug-only count hoisted into guard**

The `affectedCount` computation (iterating all dependents to count how many are affected) was performed unconditionally but only consumed inside a `logger.isDebugEnabled()` guard. Moved the entire loop inside the guard.

**Signature cleanup**

The `hasFilteredResourcesWithChangedProperty` → `scanDirectoryForPropertyRefs` → `checkFileForPropertyRefs` call chain now passes `Set<String>` property names directly instead of pre-building a `List<String>` of `\${prop}` formatted refs.

### Note on readPomText and Xpp3Dom.toString()

The issue mentions `readPomText` and an `Xpp3Dom.toString()` in a per-property loop. Both were removed by the #131 effective-model rework (commit 3fe6b72). The equivalent performance issue now lives in the filtered-resource scanning path, which this PR addresses.

### Tests

- 9 new unit tests for `containsAnyPropertyRef` covering: single match, one-of-many, no match, empty content, empty property set, multiple placeholders, unterminated placeholder, nested dollar signs, empty placeholder name, adjacent placeholders.
- All existing tests pass (no behavioral changes).

## Summary by CodeRabbit

- **Performance**
  - Improved project change analysis when processing files containing many property references.
  - Reduced unnecessary diagnostic processing when detailed debugging is disabled.

- **Reliability**
  - Expanded coverage for property-reference detection, including multiple, adjacent, empty, nested, and incomplete placeholders.

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
@@ -1561,6 +1561,7 @@ class PomChangeAnalyzer {
      * Complexity: O(len + P) where len is content length and P is |propertyNames|,
      * compared to the previous O(P × len) approach of calling {@code contains()} per property.
      */
+    @SuppressWarnings("java:S135") // two breaks are clearer than a convoluted loop condition
     static boolean containsAnyPropertyRef(String content, Set<String> propertyNames) {
         int i = 0;
         while (true) {
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
@@ -647,13 +647,13 @@ class PomChangeAnalyzer {
             }
         }
 
-        int affectedCount = 0;
-        for (MavenProject dep : dependentProjects) {
-            if (ctx.affected.contains(dep)) {
-                affectedCount++;
-            }
-        }
         if (logger.isDebugEnabled()) {
+            int affectedCount = 0;
+            for (MavenProject dep : dependentProjects) {
+                if (ctx.affected.contains(dep)) {
+                    affectedCount++;
+                }
+            }
             logger.debug(
                     "Parent {} analysis complete: {} of {} dependents affected",
                     key(parentProject),
@@ -1359,12 +1359,7 @@ class PomChangeAnalyzer {
             return false;
         }
 
-        List<String> refs = new ArrayList<>();
-        for (String prop : unscanned) {
-            refs.add("${" + prop + "}");
-        }
-
-        if (scanFilteredResources(project, refs, ctx)) {
+        if (scanFilteredResources(project, unscanned, ctx)) {
             return true;
         }
         // No match found — record scanned properties for memoization (#114)
@@ -1394,7 +1389,7 @@ class PomChangeAnalyzer {
      * Scan all filtered resource directories of the given project for any of the given
      * property references.
      */
-    private boolean scanFilteredResources(MavenProject project, List<String> refs, AnalysisContext ctx) {
+    private boolean scanFilteredResources(MavenProject project, Set<String> changedPropertyNames, AnalysisContext ctx) {
         List<Resource> allResources = new ArrayList<>();
         if (project.getResources() != null) {
             allResources.addAll(project.getResources());
@@ -1418,7 +1413,7 @@ class PomChangeAnalyzer {
             if (!Files.isDirectory(resourceDir)) {
                 continue;
             }
-            if (scanDirectoryForPropertyRefs(resourceDir, refs, ctx)) {
+            if (scanDirectoryForPropertyRefs(resourceDir, changedPropertyNames, ctx)) {
                 logger.debug(
                         "Found property reference in filtered resources of {} (dir={})", key(project), resourceDir);
                 return true;
@@ -1430,7 +1425,7 @@ class PomChangeAnalyzer {
     private static final int MAX_RESOURCE_WALK_DEPTH = 32;
     private static final int MAX_RESOURCE_WALK_FILES = 10_000;
 
-    private boolean scanDirectoryForPropertyRefs(Path dir, List<String> refs, AnalysisContext ctx) {
+    private boolean scanDirectoryForPropertyRefs(Path dir, Set<String> changedPropertyNames, AnalysisContext ctx) {
         // Does not follow symbolic links: a symlink could loop forever or point
         // outside the module (leaking file content into the analysis)
         // Every exit path adds the entries it visited to ctx.resourcesVisited (#99).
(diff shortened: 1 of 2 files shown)


extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
```diff
@@ -1496,6 +1496,33 @@ private boolean checkFileForPropertyRefs(Path entry, List<String> refs) {
...
+     * Single-pass scan of {@code content} for Maven property placeholders {@code ${name}}.
+     * Extracts every placeholder token and checks if any appears in {@code propertyNames}.
+     * <p>
+     * Complexity: O(len + P) where len is content length and P is |propertyNames|,
+     * compared to the previous O(P × len) approach of calling {@code contains()} per property.
+     */
+    static boolean containsAnyPropertyRef(String content, Set<String> propertyNames) {
+        int len = content.length();
+        int i = 0;
+        while (i < len - 2) { // need at least "${}"
```
gnodet: 💡 **Nit — the `i < len - 2` loop guard is functionally redundant.** Both `indexOf` calls already return `-1` when the substring isn't found, and the `break` on negative return handles all early exits. The guard looks like it's preventing an underflow but `len - 2` is a plain `int` subtraction (no underflow for zero-length strings). The two `break` statements are the actual termination conditions.

Consider making the termination logic self-evident:

```suggestion
        while (true) {
```

(Then the two existing `if (... < 0) { break; }` blocks are the clearly-stated exit conditions. The existing tests cover `len=0` and `len=7` with unterminated placeholder.)

gnodet (commented):
**Summary:** Solid optimization — the O(P×len) → O(len+P) replacement is correct, the `affectedCount` guard fix is a clear win, and the single-pass scanner has good coverage. One boundary behavior worth documenting.

---

**`while (i < len - 2)` guard is misleading but not wrong**

The intent is "skip entry if content is too short to possibly contain `${}`". The condition `i < len - 2` means the loop body only runs when the current position can still have at least 3 characters ahead. However, since `indexOf("${", i)` and `indexOf('}', start+2)` both return `-1` on no-match (and the `break` on `< 0` handles that), the guard is an early-exit optimization, not a correctness requirement. The loop would behave identically with `while (i < len)` or even `while (true)` (with the two breaks).

This is fine functionally, but it creates a slight readability trap: the `len - 2` could be confused with a different boundary intent. Consider a comment or replacing with `while (true)` + the existing breaks, which makes the termination conditions explicit:

```java
// Uses indexOf for termination: breaks on -1 return, no explicit length guard needed
while (true) {
    int start = content.indexOf("${", i);
    if (start < 0) break;
    int end = content.indexOf('}', start + 2);
    if (end < 0) break;
    ...
    i = end + 1;
}
```

Not a required change — but the current form looks like it's guarding against integer underflow when `len < 2`, which is an int computation (no underflow possible) and the test suite already covers `len=0` via `containsAnyPropertyRef_noMatchOnEmptyContent`.

---

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
gnodet: Fixed in 762d37159df3c44e18876a0d1e27bffc61ec5ff2. Replaced `while (i < len - 2)` with `while (true)` — the two `indexOf` + `break` pairs are the sole termination conditions now.

gnodet (commented):
**Re-review of commit 762d371**

Previous finding addressed: ✅

The `while (i < len - 2)` guard and the now-redundant `len` variable have been dropped. The replacement `while (true)` makes the two explicit `break` statements the sole, self-evident termination conditions — exactly the intent of the original suggestion. The change is minimal and correct.

No new issues in the PR diff. ast-grep broad-catch findings are pre-existing (lines 468, 1745), not introduced by this PR. SonarCloud quality gate passed. Ready to merge.

gnodet (commented):
**Re-review of commit 8be14c5**

Previous finding: ✅ (confirmed addressed in prior re-review at 762d371)

Single change in this commit: `@SuppressWarnings("java:S135")` added to `containsAnyPropertyRef`. SonarCloud's java:S135 flags the `while (true)` + two `break` idiom as "multiple break/continue in loop". The suppression is correctly scoped (method-level, not class-level), and the inline comment `// two breaks are clearer than a convoluted loop condition` provides the justification. This is the right trade-off — the previous review explicitly recommended `while (true)` + breaks over a synthetic loop condition, and silencing the mechanical rule is the appropriate follow-up.

SonarCloud reports 1 new issue in the PR analysis — this is the "Docstring coverage" check-level warning flagged by CodeRabbit (29% vs 80% threshold), not a code defect. The missing Javadoc is on test methods, which don't require it.

ast-grep broad-catch findings at lines 468 and 1746 are pre-existing — not introduced by this PR.

No new issues. Ready to merge.

gnodet (commented):
**Re-review of PR 167 after rebase (HEAD: 98d289b)**

This is a rebase-triggered re-review. The branch was force-pushed but the PR content is identical to what was reviewed at `8be14c5` — the diff between the two SHAs touches only `ScalpelReportTest.java`, which is outside this PR's change set.

**Previous findings:** ✅ All addressed (confirmed in prior re-reviews)
- `while (i < len - 2)` → `while (true)` — addressed
- `@SuppressWarnings("java:S135")` scoped at method level — addressed

**New commits:** None that affect this PR's files. The rebase pulled in upstream changes to unrelated files.

**ast-grep findings:** Two `broad-exception-catch` hits at lines 468 and 1777 of `PomChangeAnalyzer.java` — both pre-existing, not introduced by this PR.

**SonarCloud:** Quality gate passed, 0 new issues, 96.8% coverage on new code.

No new issues. Ready to merge.

PR #168, opened 2026-09-07
Fix #115: replace O(S²×T×D) fixpoint with O(M×D) worklist BFS for test-jar softening

### LINKED ISSUE #115: [perf] Test-jar softening fixpoint is O(S^2 x T x D) and hides a concurrent-modification hazard
## Problem

The test-jar softening loop restarts a full scan whenever any producer is softened: iterates all skipped projects times all compiling-test projects, with an O(D) `hasTestJarDependency` inside. Worst case O(S² x T x D), and in skip-tests mode on a 500-module reactor with one changed module, S is about 499.

There is also a latent hazard: the loop iterates a collection while its body adds to it, safe only because of an immediately-following `break`; removing or moving that break yields a `ConcurrentModificationException`.

Important context: this code implements the #47 fix, the highest-value correctness fix recently shipped. The rewrite must keep behavior identical (existing unit tests + the skip-tests-test-jar IT are the proof).

## Proposal

Build a reverse index from artifact coordinates to test-jar consumers once, O(M x D), then a worklist BFS seeded with the test projects: total O(M x D), one pass, and the CME hazard is eliminated structurally rather than by a break.

gnodet:
## Problem

The test-jar softening loop restarted a full scan whenever any producer was softened: iterating all skipped projects × all compiling-test projects, with an O(D) `hasTestJarDependency` inside. Worst case O(S² × T × D), and in skip-tests mode on a 500-module reactor with one changed module, S ≈ 499.

There was also a latent CME hazard: `compilingTests` was mutated during iteration, safe only because of an immediately-following `break`; removing or moving that break would throw `ConcurrentModificationException`.

## Fix

- Build a `Map<GA, MavenProject>` index of skipped projects for O(1) producer lookup
- Worklist BFS seeded from `testProjects`: each consumer's test-jar deps are scanned once, softened producers are enqueued, total O(M × D)
- CME hazard eliminated structurally — no collection is mutated during iteration

## Verification

- All 190 unit tests pass (including the chained fixpoint test `softensChainedTestJarProducersViaFixpointIteration`, the out-of-build-set test `softensTestJarProducerOutsideBuildSet`, and the direct consumer test)
- The `skip-tests-test-jar` integration test passes
- Behavior is identical to the previous implementation

Fixes #115

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
@@ -25,6 +25,7 @@ import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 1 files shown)


gnodet (commented):
**Summary:** Correct BFS replacement — the O(M×D) algorithm is sound, the CME hazard is eliminated, and chained propagation is preserved. One minor correctness observation and one dead-code note below.

---

**`visited.add(producer)` is dead code as a guard**

The `softenedProjects.contains(producer)` check on line 1151 already prevents re-processing any producer. By the time `visited.add(producer)` is executed (line 1165), the producer is about to be added to `softenedProjects` in the same iteration. Since `skippedProjects` and `testProjects` are disjoint by construction (the only seeds in `visited` are `testProjects`, and `skippedByGa` is built exclusively from `skippedProjects`), `visited.add(producer)` will always return `true` when reached. The `visited` set is functionally redundant — `softenedProjects` already provides the re-enqueue guard.

This is harmless and the loop terminates correctly either way, but it makes the code look like `visited` is the primary guard when `softenedProjects` is. Consider dropping `visited` and relying on `softenedProjects.contains(producer)` alone, which already appears two lines above. If the intent is to future-proof against non-disjoint inputs, add a comment explaining why both are needed.

---

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -1118,42 +1119,56 @@ private void applySkipTests(
...
+                softenedProjects.add(producer);
+                if (logger.isDebugEnabled()) {
+                    logger.debug(
+                            "Scalpel: Keeping test-compile for {} because its test-jar is consumed"
+                                    + " in-reactor by {} (softened: skipTests=true instead of"
+                                    + " maven.test.skip=true)",
+                            key(producer),
+                            key(consumer));
+                }
+                // Enqueue the newly softened producer so its own test-jar deps are checked.
```
gnodet, line 1164: 💡 **Nit — `visited.add(producer)` is always `true` here and is dead code as a guard.** `softenedProjects` is the actual re-processing guard (checked two lines above). Since `skippedByGa` and `testProjects` are disjoint sets, no entry from `skippedByGa` can already be in `visited` (which is seeded from `testProjects` only). The `if` branch never skips.

If you want to be explicit about the termination guarantee, the simpler form is:

```suggestion
                worklist.add(producer);
```

(Drop the `if` guard — `producer` was never in `visited` when it reaches this point. If disjointness invariant could ever break, an assertion is more appropriate than a silent skip.)

gnodet: Fixed in d37c762. Dropped the `visited` set entirely — `softenedProjects.contains()` is the sole re-processing guard, and the disjointness of `testProjects`/`skippedProjects` is now documented inline.

gnodet (commented):
**Re-review of commit d37c762**

Previous finding addressed: ✅

The `visited` set and the `if (visited.add(producer))` guard have been dropped entirely. `softenedProjects.contains(producer)` is now unambiguously the sole re-processing guard, exactly as suggested. The explanatory comment documenting the disjointness invariant (`testProjects` ∩ `skippedProjects` = ∅ by construction) makes the termination argument self-contained without the noise of a redundant data structure.

No new issues in the PR diff. ast-grep broad-catch findings are pre-existing (lines 90, 268, 665, 1652, 1675 — all outside the modified range). SonarCloud quality gate passed. Ready to merge.

PR #169, opened 2026-09-07
Fix #114: single-read filtered-resource scan with cross-parent memoization

### LINKED ISSUE #114: [perf] Filtered-resource scanning: three filesystem round trips per file and one full scan per changed property
## Problem

In the filtered-resource scan path: `Files.size`, then `isBinaryFile` opens the file and reads 8000 bytes, then `Files.readAllBytes` opens it a second time: three round trips per resource file where one suffices. The content is then scanned once per changed property, O(P x fileSize). The directory walk issues two extra stat syscalls per entry (`isDirectory` + `isRegularFile`) that `walkFileTree` provides for free. And there is no cross-parent memoization, so two changed parent POMs each trigger a full re-walk of the same child resource trees.

Worth doing together with the symlink/traversal hardening item since both rewrite the same walk.
Related: #83.

## Proposal

- Read once into a byte array, NUL-check the first 8000 bytes of that buffer, then a single multi-pattern pass (or reuse placeholder-token extraction) over the bytes.
- `Files.walkFileTree` instead of the manual stack.
- Two-pass structure so the union of changed properties is known before any scan, and each module's resource tree is scanned at most once per build.

gnodet:
Fixes #114

### Single-read per file
The filtered-resource scan previously performed three filesystem round trips per file:
1. `Files.size()` — stat
2. `isBinaryFile()` — open + read first 8KB
3. `Files.readAllBytes()` — open + read entire file

Now uses `BasicFileAttributes.size()` from `walkFileTree` (free) and reads the file once with `Files.readAllBytes()`. Binary detection (NUL-check on first 8000 bytes) is performed in-memory on the already-read byte array. The standalone `isBinaryFile()` method is removed.

### Cross-parent memoization
When multiple parent POMs change, the same child module resource tree was scanned once per parent. Now tracks which properties have already been scanned per child via `resourceScannedProperties` in `AnalysisContext`, so each resource directory is visited at most once per build when the effective property diffs overlap.

### Tests added
- `analyzeChanges_singleReadOptimization_binaryFileSkipped` — verifies binary files (with NUL bytes) are still correctly skipped
- `analyzeChanges_singleReadOptimization_textFileDetectsPropertyRef` — verifies text files with property references are still correctly detected
- `analyzeChanges_crossParentMemoization_sharedChildScannedOnce` — verifies that a shared child's resource tree is only scanned once when two parents change with overlapping property diffs

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
@@ -1354,13 +1354,7 @@ class PomChangeAnalyzer {
     private boolean hasFilteredResourcesWithChangedProperty(
             MavenProject project, Set<String> changedProperties, AnalysisContext ctx) {
         // Determine which properties still need scanning (cross-parent memoization)
-        Set<String> alreadyScanned = ctx.resourceScannedProperties.getOrDefault(project, Set.of());
-        Set<String> unscanned = new LinkedHashSet<>();
-        for (String prop : changedProperties) {
-            if (!alreadyScanned.contains(prop)) {
-                unscanned.add(prop);
-            }
-        }
+        Set<String> unscanned = filterUnscannedProperties(project, changedProperties, ctx);
         if (unscanned.isEmpty()) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


gnodet (commented):
Solid. Both optimizations are well-implemented and properly tested.

- **Single-read optimization**: `attrs.size()` is free from the `walkFileTree` callback, the `readAllBytes` + in-memory NUL scan is cleaner than the previous double-open approach, and the code correctly skips symlinks before calling `checkFileForPropertyRefs`. The 1 MB size guard is pre-existing, not introduced here.

- **Cross-parent memoization**: recording only on "not found" paths is exactly right. When a match is found, the child is marked affected and the outer loop's `if (ctx.affected.contains(child)) continue` prevents re-scanning — no memoization entry needed for the true case. The `resourcesVisited <= 2` assertion in the memoization test is tight and meaningful.

- SonarCloud reports 4 new issues but the quality gate passes — those appear to be test-code issues on the new `PomChangeAnalyzerTest` methods.

No changes needed from me.

---

gnodet (commented):
Re-review after new commit `fb058583` ("Fix SonarCloud issues").

**New commit is clean.** The refactoring extracts `filterUnscannedProperties()` and `scanFilteredResources()` out of `hasFilteredResourcesWithChangedProperty()` to reduce cognitive complexity. Critical correctness check: the memoization recording (`ctx.resourceScannedProperties.computeIfAbsent(project, k -> new LinkedHashSet<>()).addAll(unscanned)`) correctly stays in the caller on the no-match path — `unscanned` is invisible to `scanFilteredResources`, which is exactly right. The redundant `(int)` cast removal and test comment rephrases are trivial.

All previous findings remain confirmed clean. SonarCloud: 0 new issues · 100% coverage on new code · 0% duplication. Ready to merge.

---

PR #170, opened 2026-09-07
Fix #96: Add escaping tests and replace substring JSON extraction with parsed assertions

### LINKED ISSUE #96: [ci] Report JSON has no golden-file or escaping tests; JSON asserted by substring, never parsed
## Problem

The JSON report is a public contract consumed by real CI, and it has no golden-file test, no schema assertion and no escaping test.

Two specific fragilities:

1. **JSON is asserted by substring, never parsed.** The main report test extracts module blocks with `indexOf`/`lastIndexOf` on brace characters. This works only because module objects currently contain no nested objects; the first nested field added to the schema silently breaks ~25 report assertions, confusingly rather than at the schema. One IT hand-slices JSON the same way while five others correctly use `JsonSlurper`.
2. **Escaping is correct but untested.** The JSON escaper handles quote, backslash, newlines and control characters, but none of the 15+ `toJson` tests exercise any of them, so a future refactor can silently regress it.

## Proposal

- Golden-file test comparing `toJson()` against a checked-in fixture.
- Frozen field-set assertion so an addition forces a deliberate update.
- Escaping tests for quotes, backslashes, newlines and control characters in module names/paths.
- Replace substring extraction with a real JSON parser in tests.

gnodet:
## Summary

Fixes #96 — the report JSON contract had no escaping tests and test assertions relied on fragile substring extraction (`indexOf`/`lastIndexOf`) that would silently break with nested objects.

### Escaping tests (core — `ScalpelReportTest`)
- **`toJson_escapesDoubleQuotes`** — quotes in `changedFiles` round-trip through the JSON parser
- **`toJson_escapesBackslashes`** — backslashes survive serialization/parsing
- **`toJson_escapesNewlinesAndCarriageReturns`** — literal newlines/CRs never appear in the JSON output; `\\n`/`\\r` escapes are verified both in raw output and by round-tripping
- **`toJson_escapesTabsAndControlCharacters`** — tabs produce `\\t`, BEL (0x07) produces `\\u0007`
- **`toJson_escapesSpecialCharsInModuleFields`** — compound test exercising quotes, backslashes, newlines and tabs in `baseBranch`, `groupId`, `artifactId`, and `path` fields

### Parsed JSON assertions (extension3 — `ScalpelLifecycleParticipantTest`)
Replaced all substring-based JSON helpers with proper parsed JSON navigation:
- `extractSection`, `extractModuleBlock` (indexOf/lastIndexOf on braces) → removed
- `modulePresent`, `moduleHasReason`, `moduleHasSourceSet`, `moduleHasField`, `moduleHasAnySourceSet`, `skippedModulePresent`, `skippedModuleHasReason` → rewritten to use `findAffectedModule`/`findSkippedModule` which parse the JSON and search the typed structure
- Evidence assertions now check the parsed `List<String>` instead of doing substring matches on hand-sliced blocks
- Added a compact `JsonReader` (mirrors core's `ScalpelReportSchemaTest.Json`)

### What was already in place
The golden-file test (`toJson_matchesGoldenFile`) and frozen field-set assertion (`schemaTopLevelProperties_matchEmittableFields`) already existed on `main` — this PR completes the remaining items from #96.

### DIFF
--- a/core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelReportTest.java
+++ b/core/src/test/java/eu/maveniverse/maven/scalpel/core/ScalpelReportTest.java
@@ -481,7 +481,7 @@ class ScalpelReportTest {
     }
 
     @Test
-    void toJson_escapesSpecialCharacters() {
+    void toJson_escapesDoubleQuotes() {
         ScalpelReport report = ScalpelReport.builder()
                 .baseBranch("origin/main")
                 .fullBuildTriggered(false)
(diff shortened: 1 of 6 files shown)


paoloantinori:
This covers the issue fully on the unit side, including the raw-output checks. One item from the issue text is still outside it: the `skip-tests-downstream-exclusion-report` IT still hand-slices module objects with `lastIndexOf('{')`/`indexOf('}')`, the exact brace-counting fragility the issue names. I have that verify script converted to `JsonSlurper` (assertions per module via `find { it.artifactId == ... }`, like the other report ITs) on a local branch; happy to send it as a tiny follow-up PR on top of this, or drop the snippet here if you prefer.

gnodet:
Good catch on the IT side — that `skip-tests-downstream-exclusion-report` verify script does have the same brace-counting fragility. A follow-up PR with `JsonSlurper` assertions sounds great, happy to review it.

gnodet (commented):
**Summary:** Solid test work. The escaping tests cover all the meaningful JSON encoding cases (quotes, backslashes, newlines, control characters, compound fields). The substring extraction replacements in `ScalpelLifecycleParticipantTest` are a strict improvement — typed field access vs. hand-sliced JSON strings. The `JsonReader` implementation is correct for the report format it targets.

No issues found — this is ready to merge. The note about the IT verify script (`skip-tests-downstream-exclusion-report` still using `lastIndexOf('{')`/`indexOf('}')`) is already tracked in the thread between @paoloantinori and @gnodet — this PR correctly scopes itself to the unit test layer.

gnodet (commented):
Re-review after new commit `cfb38b7` ("Fix #96: Add escaping tests and replace substring JSON extraction with parsed assertions").

**New commit is clean** — closes the remaining IT-side fragility from #96. The four `verify.groovy` scripts are converted from hand-sliced string extraction to `JsonSlurper` parsed assertions, exactly as @paoloantinori proposed in the thread.

Semantic correctness of the IT changes:
- `shadow-failure`: `assert shadow.contains('module-b')` tightened to `shadow.wouldHaveBuilt.any { it.contains('module-b') }` — more precise, correct for this fixture
- `shadow-mode`: full typed field access (`shadow.mode == 'shadow'`, `shadow.wouldHaveSkipped.any { ... }`) — correct
- `shadow-includepaths`: correct field-level access for `wouldHaveSkipped`/`wouldHaveBuilt`
- `skip-tests-downstream-exclusion-report`: `moduleC.testsSkippedReason == null` uses Groovy's null-safe property access — correctly asserts absence of the field

The `assertNotNull` import added to `ScalpelLifecycleParticipantTest` is consistent with the new `assertNotNull(module, ...)` usages that replace the `block != null &&` pattern. SonarCloud: 0 new issues · 0% duplication. All of #96 is now covered. Ready to merge.

---

PR #171, opened 2026-09-07
Fix #113: replace linear module scan with hash-based lookup

### LINKED ISSUE #113: [perf] Module lookup is a linear scan with per-file path normalization: O(F x M) normalizations on large PRs
## Problem

`ModuleMapper`'s inner loop computes the relative path of every project (File-to-Path conversion, two `toAbsolutePath().normalize()` calls, a `relativize`, a `toString`) for every changed file; the sort comparator calls it another O(M log M) times. Because the project list is ordered longest-path-first, a file in a shallow module scans past every deeper module. At F=5000 changed files and M=500 modules that is up to 2.5M normalizations, each allocating several objects. Likely the largest pure-CPU item on a large-PR run.

The same `relativePath` computation is repeated in several `removeIf` predicates in the participant, and reactor-root normalization sits inside loops in three files.

## Proposal

- Precompute the relative path per project once, build a map keyed by module directory, then for each changed file walk its parent directories from deepest to root doing hash lookups: O(M) setup plus O(F x depth).
- Hoist reactor-root normalization out of all loops.
- The length-descending sort is a non-obvious proxy for nesting depth; sort by name count or document the invariant.

gnodet:
## Summary

Fixes #113: Module lookup was a linear scan with per-file path normalization, causing O(F × M) normalizations on large PRs.

### ModuleMapper (core fix)
- **Before:** For each changed file, linear scan through all projects computing `getRelativePath()` (two `toAbsolutePath().normalize()` calls, a `relativize`, a `toString`). Sort comparator added O(M log M) more normalization calls. At F=5000 files × M=500 modules = 2.5M normalizations.
- **After:** Precompute relative paths once per project, build a `HashMap<String, MavenProject>` keyed by module directory. For each changed file, walk parent directories doing O(1) hash lookups. **Complexity: O(M) setup + O(F × depth) lookups** — typically depth ≤ 5, making this effectively O(F).

### ScalpelLifecycleParticipant (hoist normalization)
- Hoist `reactorRoot.toAbsolutePath().normalize()` into a single `normalizedRoot` computed once at the top of `afterProjectsRead`, replacing per-iteration normalization across:
  - POM path collection loop
  - `writeReport`, `matchesIncludePaths`, `writeImpactedLog`, `applySkipTests`, `moduleKey` lambda
  - `relativePath()` method now expects a pre-normalized root

### PomChangeAnalyzer (hoist normalization)
- Same pattern: normalize `reactorRoot` once in `analyzeChanges()`, store as `normalizedRoot` in the `AnalysisContext`
- Eliminated per-call normalization in `analyzeParentPomChange` and `hasEffectiveChanges`

### Tests
- Nested module depth resolution (grandchild gets deepest match)
- Intermediate module mapping (parent module, not child)
- Performance: 200 modules × 2000 files completes in <1s
- Non-module directory fallback to root project
- `explain=false` suppresses triggering file collection
- `getRelativePath` with pre-normalized root

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
@@ -75,6 +75,7 @@ class ModuleMapper {
         // instead of a linear scan: O(M) setup + O(F × depth) lookups.
         Path rootDir = reactorRoot.toAbsolutePath().normalize();
         Map<String, MavenProject> moduleByDir = new HashMap<>();
+        Map<MavenProject, String> pathByProject = new HashMap<>();
         MavenProject rootProject = null;
         for (MavenProject project : projects) {
             String projectPath = getRelativePath(project, rootDir);
@@ -82,13 +83,14 @@ class ModuleMapper {
                 rootProject = project;
             } else {
                 moduleByDir.put(projectPath, project);
+                pathByProject.put(project, projectPath);
             }
         }
 
         for (String changedFile : changedFiles) {
             MavenProject matched = findOwningModule(changedFile, moduleByDir, rootProject);
             if (matched != null) {
-                String projectPath = matched == rootProject ? "" : getRelativePath(matched, rootDir);
+                String projectPath = matched == rootProject ? "" : pathByProject.get(matched);
                 boolean isTest = isTestPath(changedFile, projectPath);
                 Boolean existing = hasMainChange.get(matched);
                 if (existing == null) {
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
@@ -8,8 +8,7 @@
 package eu.maveniverse.maven.scalpel.extension3.internal;
 
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Comparator;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
```diff
@@ -70,29 +69,37 @@ public Result mapToProjectsClassified(
...
-                    }
-                    if (explain) {
-                        triggeringFiles
-                                .computeIfAbsent(project, k -> new LinkedHashSet<>())
-                                .add(changedFile);
-                    }
-                    break;
+            MavenProject matched = findOwningModule(changedFile, moduleByDir, rootProject);
+            if (matched != null) {
+                String projectPath = matched == rootProject ? "" : getRelativePath(matched, rootDir);
```
gnodet: 🔴 **Performance gap in the hot loop:** `getRelativePath(matched, rootDir)` is called once per matched file, doing `project.getBasedir().toPath().toAbsolutePath().normalize()` plus a `relativize()` every time. The setup phase already computed this string as the key to build `moduleByDir` — but that map stores only the project, not the path. For F=2000 files with M=200 modules this adds 2000 extra normalization calls, partially undermining the O(M) setup goal.

Keep a `Map<MavenProject, String> pathByProject` built alongside `moduleByDir` during setup and look up the path there instead:

```suggestion
                String projectPath = matched == rootProject ? "" : pathByProject.get(matched);
```

And in the setup loop:
```java
Map<String, MavenProject> moduleByDir = new HashMap<>();
Map<MavenProject, String> pathByProject = new HashMap<>();
...
    if (projectPath.isEmpty()) {
        rootProject = project;
    } else {
        moduleByDir.put(projectPath, project);
        pathByProject.put(project, projectPath);
    }
```
This makes the per-file lookup fully O(1) with no additional filesystem ops.

gnodet (commented):
Solid performance fix. The algorithm is correct, complexity analysis checks out, and the test coverage is good. One efficiency gap in the hot loop worth addressing.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ModuleMapper.java
gnodet: Fixed in 2f1c593.

PR #173, opened 2026-09-07
Fix #116: replace per-project DFS with precomputed BFS traversals

### LINKED ISSUE #116: [perf] ReactorTrimmer does one uncached DFS graph query per project; sync-point modules make it quadratic
## Problem

`DefaultProjectDependencyGraph.getDownstreamProjects(p, true)` / `getUpstreamProjects(p, true)` each perform a fresh recursive DFS plus a comparator sort, uncached. The alsoMake loop runs one such query per project in the build set, which can be all M (the code's own comment notes a sync-point module like camel-allcomponents pulls the entire reactor as upstream). So O(M x (E + k log k)), degrading to O(M² log M) in the sync-point case.

Related: `isDescendantOf` scans all M projects per changed POM and rebuilds coordinate key strings per ancestor level; a version-bump PR making C = M gives O(M² x depth).

## Proposal

- One multi-source BFS over reverse/forward edge sets built once, instead of per-project queries.
- Parent-to-children adjacency map built once for descendant computation; cached coordinate keys.
- Trim results must be identical across the existing test suite.

These would land best after the operation-count instrumentation exists so improvements are provable.

gnodet:
## Problem

`ReactorTrimmer.computeBuildSet` called `graph.getDownstreamProjects(p, true)` / `graph.getUpstreamProjects(p, true)` once per project in the build set. Each call in `DefaultProjectDependencyGraph` performs a fresh uncached DFS, giving O(M × (E + k log k)) total work, degrading to O(M² log M) when sync-point modules pull the entire reactor.

`PomChangeAnalyzer.isDescendantOf` scanned all M projects per changed POM and rebuilt coordinate key strings at each ancestor level — O(M² × depth).

### ReactorTrimmer
- Build forward (downstream) and reverse (upstream) adjacency maps **once** from direct (non-transitive) edges via `graph.getDownstreamProjects(p, false)` / `graph.getUpstreamProjects(p, false)`
- Use **multi-source BFS** over these maps to compute the transitive downstream and upstream closures in a single pass
- Total work: O(M + E) instead of O(M × DFS)

### PomChangeAnalyzer
- Build a **parent→descendants map once** via BFS over a parent→direct-children adjacency map with cached GA keys
- Replaces `findChildren()` + `isDescendantOf()` + `findParentProjects()`: O(M) total instead of O(M² × depth)

## Tests

All 193 existing tests pass — trim results are identical. 3 new tests verify:
- Multi-hop transitive downstream chain (A→B→C→D)
- Multi-hop transitive upstream chain (C←B←A)
- Diamond graph traversal without duplicates (A→B,C; B,C→D)

Fixes #116

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reactor build-set calculation for transitive upstream and downstream dependencies.
  - Preserved correct handling of test-only dependencies across multiple dependency levels and converging paths.
  - Prevented duplicate project processing in diamond-shaped dependency graphs.
  - Maintained accurate build ordering and dependency classification across complex project structures.

- **Tests**
  - Added coverage for dependency traversal, test-only filtering, duplicate elimination, build ordering, and configuration-driven behavior.
  - Improved test reliability for dependency graph scenarios.

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
@@ -87,22 +87,7 @@ class ReactorTrimmer {
                 }
                 for (MavenProject ds : directDownstream.getOrDefault(project, List.of())) {
                     if (visited.add(ds)) {
-                        buildSet.add(ds);
-                        if (config.isExplain()) {
-                            addReason(buildReasons, ds, "downstream of " + key(project));
-                        }
-                        String scope = getDependencyScope(ds, project);
-                        if ("test".equals(scope)) {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Adding test-scoped downstream {} of {}", key(ds), key(project));
-                            }
-                            downstreamTestOnly.add(ds);
-                        } else {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Adding downstream dependent {} of {}", key(ds), key(project));
-                            }
-                            downstreamOnly.add(ds);
-                        }
+                        addDownstream(ds, project, buildSet, downstreamOnly, downstreamTestOnly, buildReasons, config);
                         queue.add(ds);
                     }
                 }
@@ -112,22 +97,7 @@ class ReactorTrimmer {
                 MavenProject current = queue.poll();
                 for (MavenProject ds : directDownstream.getOrDefault(current, List.of())) {
                     if (visited.add(ds)) {
-                        buildSet.add(ds);
-                        if (config.isExplain()) {
-                            addReason(buildReasons, ds, "downstream of " + key(current));
-                        }
-                        String scope = getDependencyScope(ds, current);
-                        if ("test".equals(scope)) {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Adding test-scoped downstream {} of {}", key(ds), key(current));
-                            }
-                            downstreamTestOnly.add(ds);
-                        } else {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Adding downstream dependent {} of {}", key(ds), key(current));
-                            }
-                            downstreamOnly.add(ds);
-                        }
+                        addDownstream(ds, current, buildSet, downstreamOnly, downstreamTestOnly, buildReasons, config);
                         queue.add(ds);
                     }
                 }
@@ -156,25 +126,10 @@ class ReactorTrimmer {
                         }
                     } else {
                         visited.add(ds);
-                        buildSet.add(ds);
                         testOnlyOrigins
                                 .computeIfAbsent(ds, k -> new LinkedHashSet<>())
                                 .add(project);
-                        if (config.isExplain()) {
-                            addReason(buildReasons, ds, "downstream of " + key(project));
-                        }
-                        String scope = getDependencyScope(ds, project);
-                        if ("test".equals(scope)) {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Adding test-scoped downstream {} of {}", key(ds), key(project));
-                            }
-                            downstreamTestOnly.add(ds);
-                        } else {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Adding downstream dependent {} of {}", key(ds), key(project));
-                            }
-                            downstreamOnly.add(ds);
-                        }
+                        addDownstream(ds, project, buildSet, downstreamOnly, downstreamTestOnly, buildReasons, config);
                         queue.add(ds);
                     }
                 }
@@ -211,25 +166,10 @@ class ReactorTrimmer {
                         }
                     } else {
                         visited.add(ds);
-                        buildSet.add(ds);
                         testOnlyOrigins
                                 .computeIfAbsent(ds, k -> new LinkedHashSet<>())
                                 .addAll(currentTestOnlyOrigins);
-                        if (config.isExplain()) {
-                            addReason(buildReasons, ds, "downstream of " + key(current));
-                        }
-                        String scope = getDependencyScope(ds, current);
-                        if ("test".equals(scope)) {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Adding test-scoped downstream {} of {}", key(ds), key(current));
-                            }
-                            downstreamTestOnly.add(ds);
-                        } else {
-                            if (logger.isDebugEnabled()) {
-                                logger.debug("Adding downstream dependent {} of {}", key(ds), key(current));
-                            }
-                            downstreamOnly.add(ds);
-                        }
+                        addDownstream(ds, current, buildSet, downstreamOnly, downstreamTestOnly, buildReasons, config);
                         queue.add(ds);
                     }
                 }
@@ -280,6 +220,37 @@ class ReactorTrimmer {
         }
 
         return new TrimResult(result, directlyAffected, upstreamOnly, downstreamOnly, downstreamTestOnly, buildReasons);
+    }
+
+    /**
+     * Add a downstream project to the build set, classify its scope, log, and record the reason.
+     * Extracted to eliminate duplication across the four BFS entry points (Phase 1 seed/drain,
+     * Phase 2 seed/drain).
+     */
+    private void addDownstream(
+            MavenProject downstream,
+            MavenProject cause,
+            Set<MavenProject> buildSet,
+            Set<MavenProject> downstreamOnly,
+            Set<MavenProject> downstreamTestOnly,
+            Map<MavenProject, List<String>> buildReasons,
+            ScalpelConfiguration config) {
+        buildSet.add(downstream);
+        if (config.isExplain()) {
+            addReason(buildReasons, downstream, "downstream of " + key(cause));
+        }
+        String scope = getDependencyScope(downstream, cause);
+        if ("test".equals(scope)) {
+            if (logger.isDebugEnabled()) {
+                logger.debug("Adding test-scoped downstream {} of {}", key(downstream), key(cause));
+            }
+            downstreamTestOnly.add(downstream);
+        } else {
+            if (logger.isDebugEnabled()) {
+                logger.debug("Adding downstream dependent {} of {}", key(downstream), key(cause));
+            }
+            downstreamOnly.add(downstream);
+        }
     }
 
     private static void addReason(Map<MavenProject, List<String>> reasons, MavenProject project, String reason) {
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
@@ -10,11 +10,14 @@ package eu.maveniverse.maven.scalpel.extension3.internal;
 import static eu.maveniverse.maven.scalpel.extension3.internal.Projects.key;
 
 import eu.maveniverse.maven.scalpel.core.ScalpelConfiguration;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
 import java.util.Set;
 import javax.inject.Named;
 import javax.inject.Singleton;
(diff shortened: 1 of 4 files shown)


extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
```diff
@@ -41,71 +44,108 @@ public TrimResult computeBuildSet(
...
-                            }
+                        String scope = getDependencyScope(ds, current);
+                        if ("test".equals(scope)) {
+                            logger.debug("Adding test-scoped downstream {} of {}", key(ds), key(current));
+                            downstreamTestOnly.add(ds);
+                        } else {
+                            logger.debug("Adding downstream dependent {} of {}", key(ds), key(current));
+                            downstreamOnly.add(ds);
                         }
+                        queue.add(ds);
```
gnodet: 🔴 **Behavioral regression: test-only filtering no longer applies transitively.**

Old code: `graph.getDownstreamProjects(A, true)` returns ALL transitive downstream of A. The check `isTestOnly && !hasTestJarDependency(ds, A)` applied to every node in that transitive set — so if A is test-only, only direct test-jar consumers of A are included, and their downstream dependents are NOT (because `hasTestJarDependency(D, A)` is false when D only depends on B, not A directly).

New BFS code: test-only filtering applies only in the initial seed phase (first `for` loop over `directlyAffected`). The continuation `while (!queue.isEmpty())` block on line 105 adds downstream nodes freely, without any test-only check.

**Concrete failure scenario:** A is test-only, B depends on A via test-jar (passes filter, enters queue), D depends on B with a regular compile dep. Old code: D not included (`hasTestJarDependency(D, A)` is false). New code: D is included (B is in the queue; the while loop adds D without any check).

The comment on lines 103-104 acknowledges this change (`"non-test-only filtering applies only at the first hop"`) but that comment describes what the code does, not what it should do. The old semantics were intentional: a test-only change should not ripple transitively unless there's a test-jar chain all the way down.

**Fix options:**
1. Track which queue entries originated from a test-only source. Maintain `Map<MavenProject, Set<MavenProject>> testOnlySources` and in the while loop, check `!testOnlySources.isEmpty() && !hasTestJarDependency(ds, anyTestOnlySource)`.
2. Use a `Map<MavenProject, Boolean> fromTestOnly` alongside the queue to propagate the test-only flag, and enforce `hasTestJarDependency` at each hop for flagged entries.
3. Keep a dedicated `Queue<Pair<MavenProject, MavenProject>> testOnlyQueue` (downstream, source) processed separately.

**Missing test:** Add a test covering `A (test-only) → B (test-jar dep on A) → D (compile dep on B)` — D must NOT be in the build set.

gnodet (commented):
The BFS approach is correct for the happy path and the three new tests cover the basic topology cases. However the test-only propagation semantics changed in a way that causes over-inclusion in multi-hop chains, and there's no test covering this scenario. Everything else (upstream BFS, PomChangeAnalyzer descendant map, mock fixup) looks solid.

**This needs a fix before merge: the test-only filtering regression described in the inline comment.**

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
gnodet: Fixed in 83f1745. Propagated testOnlyOrigins through the BFS queue — at every hop from a test-only-originated node, hasTestJarDependency(ds, origin) is enforced against the original test-only source(s), matching the old DFS semantics. Added the missing test: A (test-only) → B (test-jar dep) → D (compile dep on B) — D is correctly excluded.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
```diff
@@ -41,71 +44,147 @@ public TrimResult computeBuildSet(
...
                 boolean isTestOnly = testOnlyProjects.contains(project);
-                List<MavenProject> downstream = graph.getDownstreamProjects(project, true);
-                if (!downstream.isEmpty()) {
-                    for (MavenProject ds : downstream) {
-                        if (directlyAffected.contains(ds)) {
-                            continue;
+                for (MavenProject ds : directDownstream.getOrDefault(project, List.of())) {
+                    if (visited.contains(ds)) {
+                        continue;
+                    }
```
gnodet, line 119: 🔴 **Ordering-sensitive over-filtering: non-test-only path to a node can be silently discarded.**

When `visited.contains(ds)` fires and the current `project` is non-test-only, any earlier `testOnlyOrigins[ds]` entry from a prior iteration remains. When BFS later processes `ds`, it sees `currentFromTestOnly=true` and filters `ds`'s downstream against the test-only origins — even though `ds` is reachable via a non-test-only path.

**Concrete failure scenario:**
- `directlyAffected` = {B (test-only), A (non-test-only)} (B iterated first, e.g., insertion order)
- Both B and A have direct edge to C
- C has downstream D (compile dep on C, no test-jar dep on B)
- B→C: `testOnlyOrigins[C] = {B}`, visited
- A→C: `visited.contains(C)` → continue (non-test-only path dropped silently)
- BFS processes C: `currentFromTestOnly=true`, `hasTestJarDependency(D, B)=false` → D excluded
- **Expected:** D included (C is also reachable via non-test-only A)

**Fix:** clear `testOnlyOrigins` for `ds` when a non-test-only source reaches an already-visited node:

```suggestion
                    if (visited.contains(ds)) {
                        // A non-test-only source reaching an already-visited node clears any
                        // test-only-origin tracking — the non-test-only path dominates.
                        if (!isTestOnly) {
                            testOnlyOrigins.remove(ds);
                        }
                        continue;
                    }
```

gnodet (commented):
**Re-review of commit 83f1745**

## Previous finding: test-only filtering regression

**ADDRESSED.** The `testOnlyOrigins` map correctly propagates test-only source sets through the BFS queue, and the new test (`computeBuildSet_testOnlyDoesNotPropagateTransitively`) verifies the A→B(test-jar)→D scenario. The fix matches the old DFS semantics.

## New finding: ordering-sensitive over-filtering in multi-source seed phase

The `testOnlyOrigins` approach introduces a subtle order-dependency in the seed phase that was not present in the old DFS code.

**Scenario:** `directlyAffected` = {A (non-test-only), B (test-only)}, both with a direct edge to C. C depends on A normally and on B via test-jar. C also has downstream D (compile dep on C).

- **If A is processed first** in the `for (MavenProject project : directlyAffected)` loop:
  - A→C: `visited.add(C)`, `testOnlyOrigins[C]` not set (A is not test-only)
  - B→C: `visited.contains(C)` → skip
  - BFS continuation for C: `currentFromTestOnly = false` → D propagates freely ✓

- **If B is processed first:**
  - B→C: `visited.add(C)`, `testOnlyOrigins[C] = {B}`
  - A→C: `visited.contains(C)` → skip (testOnlyOrigins[C] = {B} remains, non-test-only path is silently lost)
  - BFS continuation for C: `currentFromTestOnly = true`, checks `hasTestJarDependency(D, B)` → false → D **excluded** ✗

The old DFS code was not order-sensitive: when A (non-test-only) processed, it called `graph.getDownstreamProjects(A, true)` returning all transitive downstream including D, and since `!isTestOnly`, D was added unconditionally. B's processing was a no-op (D already in buildSet).

**The fix** is straightforward — when `visited.contains(ds)` is true in the seed phase and the current source is non-test-only, remove `ds` from `testOnlyOrigins` (the non-test-only path dominates):

```java
if (visited.contains(ds)) {
    // A non-test-only source reaching an already-visited node clears any
    // test-only-origin tracking — the non-test-only path dominates.
    if (!isTestOnly) {
        testOnlyOrigins.remove(ds);
    }
    continue;
}
```

The same applies in the `while (!queue.isEmpty())` continuation: if `currentFromTestOnly=false` and a node already in `testOnlyOrigins` is re-encountered (skipped via `visited.contains`), the test-only tracking should be cleared. Though this case is rarer since the visited check prevents re-entry.

**Missing test:** Add a test for `A (non-test-only) → C, B (test-only) → C (test-jar), C → D` — D must be in the build set regardless of the iteration order of A and B in `directlyAffected`.

---

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
gnodet: Fixed in c832518. Applied the suggested fix in both BFS loops: the initial directlyAffected seed loop (using `!isTestOnly`) and the transitive BFS loop (using `!currentFromTestOnly`). When a non-test-only source reaches an already-visited node, `testOnlyOrigins` is cleared for that node so the non-test-only path dominates. All 194 unit tests pass.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
```diff
@@ -41,71 +44,172 @@ public TrimResult computeBuildSet(
...
+                            }
+                            downstreamTestOnly.add(ds);
+                        } else {
+                            if (logger.isDebugEnabled()) {
+                                logger.debug("Adding downstream dependent {} of {}", key(ds), key(current));
+                            }
+                            downstreamOnly.add(ds);
+                        }
+                        queue.add(ds);
                     }
```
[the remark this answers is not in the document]
gnodet: Fixed in a7f7c83. Refactored into two-phase BFS: Phase 1 walks from non-test-only sources with no filtering; Phase 2 walks from test-only sources with `hasTestJarDependency` at every hop, treating Phase 1-visited nodes as safe. This eliminates the processing-order race entirely since non-test-only paths are fully resolved before test-only paths are considered. Added regression test for the converging-paths scenario (A(test-only)→X, B(non-test)→C→X, X→Y — Y must be included).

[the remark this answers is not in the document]
gnodet: Fixed in a7f7c83. Refactored into two-phase BFS: Phase 1 walks from non-test-only sources with no filtering; Phase 2 walks from test-only sources with `hasTestJarDependency` at every hop, treating Phase 1-visited nodes as safe. This eliminates the processing-order race entirely since non-test-only paths are fully resolved before test-only paths are considered. Added regression test for the converging-paths scenario (A(test-only)→X, B(non-test)→C→X, X→Y — Y must be included).

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmer.java
```diff
@@ -41,71 +44,160 @@ public TrimResult computeBuildSet(
...
+                if (!testOnlyProjects.contains(project)) {
+                    continue; // handled in Phase 1
+                }
+                for (MavenProject ds : directDownstream.getOrDefault(project, List.of())) {
+                    if (visited.contains(ds)) {
+                        // Already reached via a non-test-only path (or another test-only
+                        // source) — descendants are already queued from Phase 1, so skip.
+                        continue;
+                    }
+                    if (!hasTestJarDependency(ds, project)) {
```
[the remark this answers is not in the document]
gnodet: Not a valid finding. The scenario described (A→B→D where D has `hasTestJarDependency(D, A) = true` but A→D is not a direct reactor edge) cannot occur in a real Maven reactor. If D declares a `<dependency>` on A with `<type>test-jar</type>`, Maven's `ProjectDependencyGraph` creates a direct edge A→D. Phase 2's initial seed loop iterates `directDownstream.get(A)`, which includes D — so D is examined directly from A, not only reachable through B.

The scenario where D has no direct dependency on A (only transitive through B) means `hasTestJarDependency(D, A)` is `false` — and both old DFS and new BFS correctly exclude D.

For multiple test-only sources sharing a node: already handled by the two-phase design — Phase 1 fully resolves non-test-only reachability before Phase 2 runs, so shared nodes reached via any non-test path are safe.

gnodet (commented):
**Re-review of commits c832518 → 326ee84b**

## Previous findings

**ADDRESSED — ordering-sensitive over-filtering (c832518 + a7f7c83).**

The single-queue interleaved BFS race (a test-only path deciding a node's downstream before a later non-test-only path clears its `testOnlyOrigins` entry) is now eliminated by the two-phase split:
- Phase 1 runs to completion, fully resolving all nodes reachable via non-test-only sources before Phase 2 starts.
- Phase 2 only runs from test-only sources; nodes already in `visited` from Phase 1 are skipped unconditionally.
- The converging-path regression test (`computeBuildSet_testOnlyAndNonTestOnlyConvergingPaths`) covers the exact failure scenario previously identified.

## New commits — observations

**`a7f7c83` — two-phase BFS:** Correct. One dead-code note: the `if (currentTestOnlyOrigins == null) continue` guard at lines 143–147 in `ReactorTrimmer.java` cannot trigger after Phase 1's drain loop empties the queue — every item enqueued in Phase 2 has a `testOnlyOrigins` entry by construction. The defensive comment is accurate, but the branch is unreachable. Non-blocking; costs nothing.

**`7c5af3a` — `addDownstream` helper:** Correct. Called from all four BFS entry points with the right `cause` argument. Scope classification (`getDependencyScope(downstream, cause)`) is semantically consistent with old behavior: first-hop items use the direct scope (test-jar → `downstreamTestOnly`), transitive-hop items where `cause` is intermediate use the intermediate scope (typically compile → `downstreamOnly`).

**`16324f2` — Windows CI fix (merged from #180):** The `ChangeFilter` refactor from `PathMatcher`/`Path.of()` to `Pattern`/`globToPattern` is correct and portable. The `globToPattern` logic handles all three `**` forms correctly:
- `**/suffix` → `(.*/)?` + suffix: correctly rejects `parentpom.xml` (Java `Matcher.matches()` is a full-string match, so the optional prefix can't consume partial segments)
- `**` at end → `.*`: safe, no following literals to cause catastrophic backtracking
- `**` followed by non-`/` → `[^/]*`: single-segment fallback, ReDoS fix

The `ReactorFirstModelResolver.addRepository()` no-op is correct: reactor-local resolution needs no remote repositories, and a fully-configured `RemoteRepositoryManager` is not guaranteed during standalone effective model building.

**`buildDescendantMap`:** BFS per parent over `childrenMap`. Maven rejects circular parent inheritance at load time, so no cycle guard is needed — correct omission.

---

All previous findings addressed. New commits solid. Ready to merge.

PR #174, opened 2026-09-07
Fix #125: extract single-responsibility collaborators from ScalpelLifecycleParticipant

### LINKED ISSUE #125: [arch] ScalpelLifecycleParticipant is a 309-line method with six early returns across four concerns
## Problem

`afterProjectsRead` is a single 309-line method with six early-return points performing config bootstrap, git orchestration, glob filtering, change classification, POM analysis dispatch, force-build selection, includePaths filtering, transitive impact analysis, trim-set computation, report writing, impacted-log writing and skip-tests application. The class holds five collaborators and 22 private methods, plus a triplicated near-identical "no affected modules" block in three places.

Natural seams: glob compilation (PathFilters, separate issue), changed-file classification, transitive impact analysis with the resolver cache, skip-tests application, report assembly, impacted-log writing, forceBuild matching. Target: the participant becomes ~150 lines of orchestration.

Sequencing note: worth doing after the open correctness fixes land; refactoring around known bugs means doing the work twice.

## Proposal

Extract the listed classes with single responsibilities and constructor-injected collaborators; deduplicate the no-affected-modules block; each extracted class unit-testable without a full MavenSession mock; behavior unchanged per the existing suite.

gnodet:
## Summary

Extracts six single-responsibility collaborator classes from the 2008-line `ScalpelLifecycleParticipant`, reducing it to a 736-line orchestrator:

| Class | Responsibility | Lines |
|-------|---------------|-------|
| `ChangedFileClassifier` | Glob compilation, path filtering, disable/full-build triggers, POM/source classification, includePaths | 174 |
| `TransitiveImpactAnalyzer` | Two-pass transitive impact analysis, resolver cache, dependency tree diff, effective model comparison | 504 |
| `SkipTestsApplier` | Skip-tests mode, test-jar softening, per-category args, downstream exclusion | 268 |
| `ReportAssembler` | JSON report assembly, status/shadow/full-build report writing | 330 |
| `ImpactedLogWriter` | Impacted-log file writing with path safety validation | 122 |
| `ForceBuildMatcher` | Regex-based forceBuildModules pattern matching | 40 |

### Design

- All extracted classes are **package-private** with **constructor-injected** collaborators
- Each class is **independently unit-testable** without a full `MavenSession` mock
- The participant's DI constructor signature is **unchanged** — internal collaborators are constructed from existing injected dependencies
- The triplicated "no affected modules" block is **deduplicated** into `handleNoAffectedModules()`
- Static forwarders `isSafeImpactedLogPath()` and `normalizeGlobPattern()` preserve backward compatibility with existing tests

### Tests

- All **95 existing `ScalpelLifecycleParticipantTest`** tests pass unchanged
- **54 new tests** across 5 test classes cover the extracted logic:
  - `ChangedFileClassifierTest` (17 tests)
  - `TransitiveImpactAnalyzerTest` (12 tests)
  - `SkipTestsApplierTest` (11 tests)
  - `ImpactedLogWriterTest` (7 tests)
  - `ForceBuildMatcherTest` (7 tests)
- Build: **244 tests, 0 failures, 0 errors** ✅

### Notes

The participant is 736 lines rather than the aspirational ~150 from the issue. The remaining bulk is pure orchestration: the 530-line `afterProjectsRead` method sequences config bootstrap → change detection → classification → POM analysis → transitive resolution → includePaths filtering → shadow mode → trim/skip-tests dispatch. All domain **logic** has been extracted; what remains is glue code that wires collaborator outputs together and handles the mode-dependent branching. Further decomposition would require splitting the method into smaller orchestration phases, which would benefit from landing the shadow-mode extraction first (separate concern).

Fixes #125

## Summary by CodeRabbit

- **New Features**
  - Selective builds now account for source, POM, dependency, plugin, and transitive changes.
  - Modules can be force-built using configured patterns.
  - Test skipping preserves test-jar outputs required by other modules.
  - Impacted-module logs include validated paths and relative module names.
  - Build reports include affected, skipped, excluded, and timing details.
- **Bug Fixes**
  - Improved handling of excluded paths, include patterns, unresolved dependencies, and full-build triggers.
  - Added safer handling for invalid report and log paths.

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ChangedFileClassifier.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ChangedFileClassifier.java
@@ -1,0 +1,178 @@
+/*
+ * Copyright (c) Maveniverse Org.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+package eu.maveniverse.maven.scalpel.extension3.internal;
+
+import eu.maveniverse.maven.scalpel.core.ScalpelConfiguration;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import org.apache.maven.project.MavenProject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Classifies changed files into POM changes vs source changes, applies exclusion/inclusion
+ * filters, and detects disable/full-build triggers.
+ */
+class ChangedFileClassifier {
+
+    private static final String GLOB_PREFIX = "glob:";
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    /**
+     * Result of classifying changed files into POM changes and source changes.
+     */
+    static final class ClassificationResult {
+        final Set<String> pomChanges;
+        final Set<String> sourceChanges;
+
+        ClassificationResult(Set<String> pomChanges, Set<String> sourceChanges) {
+            this.pomChanges = pomChanges;
+            this.sourceChanges = sourceChanges;
+        }
+    }
+
+    /**
+     * Separates the given changed files into POM changes and source changes.
+     * Files under {@code .mvn/} are ignored (build infrastructure).
+     */
+    ClassificationResult classifyChanges(Set<String> changedFiles) {
+        Set<String> pomChanges = new LinkedHashSet<>();
+        Set<String> sourceChanges = new LinkedHashSet<>();
+        for (String file : changedFiles) {
+            if (file.endsWith("/pom.xml") || file.equals("pom.xml")) {
+                pomChanges.add(file);
+            } else if (file.startsWith(".mvn/")) {
+                logger.debug("Ignoring build infrastructure file: {}", file);
+            } else {
+                sourceChanges.add(file);
+            }
+        }
+        return new ClassificationResult(pomChanges, sourceChanges);
+    }
+
+    /**
+     * Returns {@code true} if any changed file matches one of the configured disable triggers.
+     */
+    boolean matchesDisableTrigger(Set<String> changedFiles, ScalpelConfiguration config) {
+        for (String pattern : config.getDisableTriggers()) {
+            PathMatcher matcher = FileSystems.getDefault().getPathMatcher(GLOB_PREFIX + normalizeGlobPattern(pattern));
+            for (String changedFile : changedFiles) {
+                if (matcher.matches(Path.of(changedFile))) {
+                    logger.info(
+                            "Scalpel: Disabled due to change in {} (matches disable trigger {})", changedFile, pattern);
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Removes files matching exclusion path patterns from the changed file set.
+     */
+    Set<String> filterExcludedPaths(Set<String> changedFiles, ScalpelConfiguration config) {
+        List<PathMatcher> excludeMatchers = compileGlobMatchers(config.getExcludePaths());
+        if (excludeMatchers.isEmpty()) {
+            return changedFiles;
+        }
+        Set<String> filtered = new LinkedHashSet<>();
+        for (String file : changedFiles) {
+            boolean excluded = false;
+            for (PathMatcher matcher : excludeMatchers) {
+                if (matcher.matches(Path.of(file))) {
+                    excluded = true;
+                    break;
+                }
+            }
+            if (!excluded) {
+                filtered.add(file);
+            }
+        }
+        int excludedCount = changedFiles.size() - filtered.size();
+        if (excludedCount > 0) {
+            logger.info("Scalpel: {} files excluded by path filters", excludedCount);
+        }
+        return filtered;
+    }
+
+    /**
+     * Returns the first changed file matching a full-build trigger pattern, or {@code null}
+     * if no trigger matches.
+     */
+    String findFullBuildTrigger(Set<String> changedFiles, ScalpelConfiguration config) {
+        for (String pattern : config.getFullBuildTriggers()) {
+            PathMatcher matcher = FileSystems.getDefault().getPathMatcher(GLOB_PREFIX + normalizeGlobPattern(pattern));
+            for (String changedFile : changedFiles) {
+                if (matcher.matches(Path.of(changedFile))) {
+                    logger.info("Scalpel: Full build triggered by change to {} (matches {})", changedFile, pattern);
+                    return changedFile;
+                }
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Returns {@code true} if the project's module path matches at least one of the
+     * given include-paths matchers.
+     */
+    static boolean matchesIncludePaths(MavenProject project, List<PathMatcher> matchers, Path reactorRoot) {
+        String relPath = relativePath(reactorRoot, project);
+        Path modulePath = Path.of(relPath);
+        for (PathMatcher matcher : matchers) {
+            if (matcher.matches(modulePath) || matcher.matches(modulePath.resolve("pom.xml"))) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Compiles a list of glob patterns into PathMatcher instances.
+     */
+    static List<PathMatcher> compileGlobMatchers(List<String> patterns) {
+        if (patterns.isEmpty()) {
+            return List.of();
+        }
+        List<PathMatcher> matchers = new ArrayList<>(patterns.size());
+        for (String pattern : patterns) {
+            matchers.add(FileSystems.getDefault().getPathMatcher(GLOB_PREFIX + normalizeGlobPattern(pattern)));
+        }
+        return matchers;
+    }
+
+    /**
+     * Normalizes a user-supplied glob pattern so that bare patterns (those containing no path
+     * separator) match files at any depth in the repository tree.
+     */
+    static String normalizeGlobPattern(String pattern) {
+        if (pattern.contains("/")) {
+            return pattern;
+        }
+        return "{" + pattern + ",**/" + pattern + "}";
+    }
+
+    /**
+     * Computes the relative path from the (pre-normalized) reactor root to the project's
+     * base directory.  Callers must pass a root that has already been normalized via
+     * {@code Path.toAbsolutePath().normalize()} — this method does NOT re-normalize the
+     * root on every call (#113).
+     */
+    static String relativePath(Path normalizedRoot, MavenProject project) {
+        return normalizedRoot
+                .relativize(project.getBasedir().toPath().toAbsolutePath().normalize())
+                .toString()
+                .replace('\\', '/');
+    }
+}
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ChangedFileClassifier.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ChangedFileClassifier.java
@@ -127,8 +127,8 @@ class ChangedFileClassifier {
      * Returns {@code true} if the project's module path matches at least one of the
      * given include-paths matchers.
      */
-    static boolean matchesIncludePaths(MavenProject project, List<PathMatcher> matchers, Path reactorRoot) {
-        String relPath = relativePath(reactorRoot, project);
+    static boolean matchesIncludePaths(MavenProject project, List<PathMatcher> matchers, Path normalizedRoot) {
+        String relPath = relativePath(normalizedRoot, project);
         Path modulePath = Path.of(relPath);
         for (PathMatcher matcher : matchers) {
             if (matcher.matches(modulePath) || matcher.matches(modulePath.resolve("pom.xml"))) {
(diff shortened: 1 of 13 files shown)


gnodet (commented):
Sound refactoring overall — all extracted classes are package-private, constructor-injected, and the deduplication of the `handleNoAffectedModules` triple into one helper is a clear improvement. The 95 existing tests pass and 54 new ones cover the extracted logic. One real issue: `ChangedFileClassifier.relativePath` re-normalizes `reactorRoot` on every call, which undoes a targeted optimization introduced in PR #171. Two nits on package-private visibility decisions. SonarCloud flagged 32 new issues; the `catch (Exception e)` in `writeStatusReport`/`writeShadowStatus` are intentional (moved verbatim from the original SLP) and not a real regression.

**The normalization issue needs addressing before merge, particularly because it affects call sites that receive `normalizedRoot` from the orchestrator.**

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ChangedFileClassifier.java
```diff
@@ -0,0 +1,174 @@
...
+        return "{" + pattern + ",**/" + pattern + "}";
+    }
+
+    static String relativePath(Path reactorRoot, MavenProject project) {
+        return reactorRoot
+                .toAbsolutePath()
+                .normalize()
+                .relativize(project.getBasedir().toPath().toAbsolutePath().normalize())
+                .toString()
+                .replace('\\', '/');
```
gnodet: 🔴 **Per-call normalization — inconsistent with #171's hoisting optimization.**

`ChangedFileClassifier.relativePath` calls `reactorRoot.toAbsolutePath().normalize()` on every invocation. PR #171 specifically hoisted this into a single `normalizedRoot` in `ScalpelLifecycleParticipant.afterProjectsRead` and passed the pre-normalized root to all internal helpers. PR #174 (built independently from the same base) extracted these methods but didn't carry that optimization forward.

The call sites in the orchestrator now pass the raw `reactorRoot` back to `ChangedFileClassifier.relativePath`, triggering redundant normalization. This matters most in `ReportAssembler.addDirectlyAffectedModules` and `addTransitivelyAffectedModules` (called once per affected module) and less so in the filter passes — but it directly contradicts the stated intent of #113.

The fix is to accept an already-normalized root. Since this is a package-private static, the signature change doesn't break any public API:

```suggestion
    static String relativePath(Path normalizedRoot, MavenProject project) {
        return normalizedRoot
                .relativize(project.getBasedir().toPath().toAbsolutePath().normalize())
                .toString()
                .replace('\\', '/');
    }
```

And the `matchesIncludePaths` method should receive the same pre-normalized root. Callers in `ScalpelLifecycleParticipant` should pass the `normalizedRoot` that was already computed at line 123.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ReportAssembler.java
```diff
@@ -0,0 +1,330 @@
...
+
+    ReportAssembler(ChangedFileClassifier changedFileClassifier) {
+        this.changedFileClassifier = changedFileClassifier;
+    }
+
+    void writeReport(
+            ScalpelConfiguration config,
+            Path reactorRoot,
+            List<MavenProject> allProjects,
+            AnalysisContext ctx,
```
gnodet, line 45: 🔵 **Nit — `handleWriteFailure` is package-private (accessible to tests) but its entire logic is duplicated in `ImpactedLogWriter`.** Both classes independently implement the same `if (config.isFailSafe()) warn else throw` pattern with the same message format. Consider a shared `static` helper in a package-private utility, or expose it on `ReportAssembler` and let `ImpactedLogWriter` call it. Neither is blocking, but it's the kind of duplication the issue description explicitly calls out as a motivation for this PR.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ChangedFileClassifier.java
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Returns {@code true} if the project's module path matches at least one of the
+     * given include-paths matchers.
+     */
+    static boolean matchesIncludePaths(MavenProject project, List<PathMatcher> matchers, Path reactorRoot) {
```
gnodet: 🔵 **Nit — `matchesIncludePaths` is `static` but accesses `reactorRoot` via `relativePath`, making it non-purely-functional.** The method is package-private, so this doesn't affect callers outside the package. But since it's used both from `ReportAssembler` and from `ScalpelLifecycleParticipant`, the fact that the root normalization is baked into each call (instead of being a pre-computed argument) will become a problem once #171 is merged and the orchestrator starts distinguishing `reactorRoot` from `normalizedRoot`. Harmonize now rather than after a rebase.

gnodet (commented):
Re-review after new commits (8dbf649 + merge of main into fix/125).

**Prior findings — status:**

✅ **Fixed**: `relativePath` re-normalization — parameter renamed to `normalizedRoot`, `toAbsolutePath().normalize()` removed from the method body. All call sites in `ScalpelLifecycleParticipant` updated to pass the pre-computed `normalizedRoot`. The `#113` optimization is now correctly preserved in the extracted collaborator.

✅ **Fixed**: `ImpactedLogWriter.handleWriteFailure` duplication — `ImpactedLogWriter` now takes a `ReportAssembler` in its constructor and delegates both failure calls to `reportAssembler.handleWriteFailure`.

✅ **Fixed (incidental)**: `matchesIncludePaths` and `relativePath` in `ScalpelLifecycleParticipant` now route through `ChangedFileClassifier.*` static methods rather than the old local private copies — the private methods are gone.

⚠️ **Three nits from the previous review remain open** (see inline comments below). None are blocking correctness, but they were explicitly flagged and not addressed.

The main refactoring itself is solid — the collaborator boundary is clean, the SLP orchestrator is significantly smaller, and the new unit tests cover the extracted logic well.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ChangedFileClassifier.java
```diff
@@ -0,0 +1,178 @@
...
```
gnodet: 🔵 **Nit (from prior review) — parameter name not updated to match the contract.**

`relativePath` was correctly renamed to `normalizedRoot`, and its Javadoc explains the pre-normalization contract. `matchesIncludePaths` passes that same pre-normalized root to `relativePath`, so callers must satisfy the same contract — but its parameter is still named `reactorRoot`, which implies a raw (unnormalized) path is acceptable.

```suggestion
    static boolean matchesIncludePaths(MavenProject project, List<PathMatcher> matchers, Path normalizedRoot) {
```

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/SkipTestsApplier.java
```diff
@@ -0,0 +1,269 @@
...
+            return false;
+        }
+        if (!matchesDownstreamExclusion(project, config.getSkipTestsForDownstreamModules())) {
+            return false;
+        }
+        // Safety guard: don't skip tests if the module has effective model changes
+        return !transitiveImpactAnalyzer.hasEffectiveModelChanges(project, models, rctx);
+    }
+
+    boolean matchesDownstreamExclusion(MavenProject project, List<String> patterns) {
```
gnodet: 🔵 **Nit (from prior review, raised by CodeRabbit) — `matchesDownstreamExclusion` is still duplicated verbatim in `ReportAssembler` (line 324).**

Both classes contain an identical method body. `Projects` already hosts the `key()` helper that both methods use. Moving `matchesDownstreamExclusion` to `Projects` eliminates the duplication and makes future changes to the matching logic a single-site edit.

Suggested move target:
```java
// In Projects.java
static boolean matchesDownstreamExclusion(MavenProject project, List<String> patterns) {
    for (String pattern : patterns) {
        if (pattern.contains(":")) {
            if (key(project).equals(pattern)) return true;
        } else {
            if (project.getArtifactId().equals(pattern)) return true;
        }
    }
    return false;
}
```
Then both `SkipTestsApplier` and `ReportAssembler` call `Projects.matchesDownstreamExclusion(project, patterns)` instead.

extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/TransitiveImpactAnalyzerTest.java
```diff
@@ -0,0 +1,240 @@
...
+        DependencyNode root = new DefaultDependencyNode((Dependency) null);
+        DependencyNode child1 = createDependencyNode("com.example", "dep-a", "1.0", "compile");
+        DependencyNode child2 = createDependencyNode("com.example", "dep-a", "2.0", "compile");
+        root.setChildren(List.of(child1, child2));
+
+        Map<String, String> versions = TransitiveImpactAnalyzer.collectDependencyVersions(root);
+
+        assertEquals(1, versions.size());
+        // First encountered wins
+        assertNotNull(versions.get("com.example:dep-a"));
```
gnodet: 🔵 **Nit (from prior review, raised by CodeRabbit) — assertion doesn't verify which version wins.**

`collectDependencyVersions` uses a LIFO stack: `child2` (`dep-a:2.0`) is pushed first, so it's visited first and its version is recorded. The assertion says `assertNotNull` but the comment says "first encountered wins" — these don't agree with each other or with the actual LIFO behavior. The test should pin the observed value:

```suggestion
        assertEquals("2.0", versions.get("com.example:dep-a"));
```

This makes the first-encountered-wins semantics observable and ensures a future change to the traversal order breaks the test rather than silently changing behavior.

extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/SkipTestsApplierTest.java
```diff
@@ -0,0 +1,216 @@
...
+                new TransitiveImpactAnalyzer.EffectiveModels(java.util.Map.of(), java.util.Map.of()),
+                new TransitiveImpactAnalyzer.ResolutionContext(
+                        tempDir,
+                        mock(org.apache.maven.execution.MavenSession.class),
+                        new java.util.LinkedHashMap<>(),
+                        new java.util.LinkedHashMap<>(),
+                        new eu.maveniverse.maven.scalpel.core.Timings())));
+    }
+
+    private MavenProject createProject(String artifactId) {
```
gnodet, line 218: 🔵 **Nit (from prior review, raised by CodeRabbit) — only the negative paths are tested.**

There is no test for the case where the pattern matches, the project is in `downstreamOnly`, and `EffectiveModels` is empty — i.e., the case that should return `true`. The two existing tests only assert `false`. The positive path (the happy path that the feature actually enables) is untested.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ChangedFileClassifier.java
gnodet: Fixed in 474df76. Renamed `reactorRoot` → `normalizedRoot` in `matchesIncludePaths` to align with the pre-normalization contract.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/SkipTestsApplier.java
gnodet: Fixed in 474df76. Moved `matchesDownstreamExclusion` to `Projects` — both `SkipTestsApplier` and `ReportAssembler` now use the shared implementation via static import.

extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/TransitiveImpactAnalyzerTest.java
gnodet: Fixed in 474df76. Assertion now pins `"2.0"` — the LIFO stack pops `child2` (dep-a:2.0) first, so that's the version that wins.

extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/SkipTestsApplierTest.java
gnodet: Fixed in 474df76. Added `shouldSkipTestsForExcludedDownstream_returnsTrue_whenMatchingDownstreamWithNoModelChanges` — covers the positive path (matching pattern + downstream project + empty EffectiveModels → returns true).

gnodet (commented):
Re-review after 474df76 — all three nits from the previous review are now addressed:

✅ `matchesIncludePaths`: parameter renamed `reactorRoot` → `normalizedRoot` to match the pre-normalization contract.

✅ `matchesDownstreamExclusion`: moved to `Projects` utility class. Both `SkipTestsApplier` and `ReportAssembler` now use the shared implementation via static import — no more duplication.

✅ `collectDependencyVersions_deduplicatesByGA`: assertion pinned to `"2.0"`, documenting the LIFO traversal order.

✅ **Bonus**: added the positive `shouldSkipTestsForExcludedDownstream` test covering the happy path (matching pattern + downstream project + empty EffectiveModels → `true`).

Clean fix-up commit, no new issues. All prior findings resolved — LGTM.

gnodet (commented):
Re-review after commit 474df76 — all four open nits from the previous review are now addressed:

✅ **`matchesIncludePaths` parameter renamed** — `reactorRoot` → `normalizedRoot`, aligning with the pre-normalization contract established by `relativePath`.

✅ **`matchesDownstreamExclusion` moved to `Projects`** — both `SkipTestsApplier` and `ReportAssembler` now use the shared implementation via static import, eliminating the verbatim duplication.

✅ **`collectDependencyVersions` assertion pinned** — `assertNotNull` replaced with `assertEquals("2.0", ...)`, making the LIFO first-encountered-wins semantics observable and regression-safe.

✅ **Positive-path test added** — `shouldSkipTestsForExcludedDownstream_returnsTrue_whenMatchingDownstreamWithNoModelChanges` covers the happy path (matching pattern + downstream project + empty EffectiveModels → returns true).

The refactoring is solid — clean collaborator boundaries, significantly smaller orchestrator, and comprehensive test coverage for the extracted logic. Unit tests pass on all platforms; verify matrix still running on the latest commit but no regressions expected.

No remaining findings — LGTM. ✅

PR #175, opened 2026-09-07
Fix #148: include/exclude glob patterns for effective model comparison

### LINKED ISSUE #148: Allow users to include/exclude specific properties from effective model comparison
## Context

PR #145 (fix for #131) introduced principled effective model comparison: only properties, managed dependencies, and managed plugins explicitly defined in a child's raw POM are compared between old and new effective models. This avoids false positives from inherited-but-unused elements while catching real changes.

There are two complementary improvements to consider:

## 1. Be smarter by default

The analyzer should understand which POM sections semantically affect a child's build output, rather than relying on a fixed list of comparisons. Some elements need special treatment:

| Element | Smart rule |
|---|---|
| `<modules>` | Adding/removing a module in the parent does NOT affect existing children's build output |
| `<properties>` | Only if the child uses the property (current: checks raw model declaration + filtered resources) |
| `<dependencyManagement>` | Only if the child consumes that GA (current: checks raw model declaration) |
| `<pluginManagement>` | Only if the child consumes that GA (current: checks raw model declaration) |
| `<dependencies>` | Always affects child (inherited) |
| `<distributionManagement>` | Does not affect compilation output |
| `<profiles>` | Only if active — and only the build-relevant sections within them |
| `<scm>`, `<developers>`, `<licenses>` | Metadata — does not affect compilation |

The current code is already partially smart (it only compares deps, plugins, properties, managed deps, managed plugins), but the logic is implicit. Making it explicit would help reason about correctness and document the guarantees.

## 2. User-configurable include/exclude as escape hatch

For cases where the smart logic can't determine impact, let users override with glob patterns over a normalized "change path" scheme:

```
# Exclude volatile properties
-Dscalpel.excludeChanges=properties/build.timestamp,properties/git.*

# Exclude all managed dependency changes (trust parent BOM blindly)  
-Dscalpel.excludeChanges=managedDependencies/**

# Force-check compiler settings even if not in raw model
-Dscalpel.includeChanges=properties/maven.compiler.*
```

### Change path scheme

| Category | Path format | Example |
|---|---|---|
| Property | `properties/<name>` | `properties/lib.version` |
| Dependency | `dependencies/<groupId>:<artifactId>` | `dependencies/com.foo:bar` |
| Plugin | `plugins/<groupId>:<artifactId>` | `plugins/org.apache.maven.plugins:maven-compiler-plugin` |
| Managed dependency | `managedDependencies/<groupId>:<artifactId>` | `managedDependencies/com.foo:bar` |
| Managed plugin | `managedPlugins/<groupId>:<artifactId>` | `managedPlugins/org.apache.maven.plugins:maven-surefire-plugin` |
| Filtered resource | `resources/<path>` | `resources/src/main/resources/app.properties` |

### Use cases for escape hatch

- **Exclude volatile properties**: `build.timestamp`, `git.commit.id`, `project.build.outputTimestamp` — change on every build, don't affect compilation
- **Exclude CI flags**: `ci.deploy.skip`, `gpg.skip` — irrelevant to build output
- **Include inherited properties**: properties that affect the build through indirect mechanisms (profile-activated plugin configuration) but aren't in the child's raw model
- **Exclude metadata dependencies**: test-only or documentation-only managed deps where version bumps shouldn't trigger rebuilds

## Design principle

Be smart by default — the include/exclude config is an escape hatch, not the primary mechanism. Users should not need to configure anything for common cases like `<modules>` changes or new BOM entries.

## Implementation notes

- The change path scheme maps naturally to the existing `EvidenceDetail` structure
- Glob matching can reuse the existing `PathMatcher` / `normalizeGlobPattern` infrastructure
- Configuration follows the established `parseList()` comma-separated pattern

gnodet:
## Summary

Fixes #148: Allow users to include/exclude specific properties from effective model comparison.

## Changes

Adds two new configuration properties:

- **`scalpel.excludeChanges`** — comma-separated glob patterns over change paths to exclude from comparison
- **`scalpel.includeChanges`** — comma-separated glob patterns that override exclusions (escape hatch)

### Change Path Scheme

| Category | Path format | Example |
|----------|------------|---------|
| Property | `properties/<name>` | `properties/build.timestamp` |
| Dependency | `dependencies/<GA>` | `dependencies/com.foo:bar` |
| Plugin | `plugins/<GA>` | `plugins/o.a.m.p:maven-compiler-plugin` |
| Managed dependency | `managedDependencies/<GA>` | `managedDependencies/com.foo:bar` |
| Managed plugin | `managedPlugins/<GA>` | `managedPlugins/o.a.m.p:maven-surefire-plugin` |

### Smart Defaults

- **Default excludes:** `properties/build.timestamp,properties/project.build.outputTimestamp` (volatile properties that change on every build)
- **`<modules>` changes** already don't trigger child rebuilds (documented)
- **Metadata sections** (scm, developers, licenses) already not compared

### Usage Examples

```bash
# Exclude git properties and CI flags
-Dscalpel.excludeChanges=properties/build.timestamp,properties/project.build.outputTimestamp,properties/git.*

# Trust parent BOM blindly
-Dscalpel.excludeChanges=managedDependencies/**

# Force-check compiler settings even when all properties are excluded
-Dscalpel.includeChanges=properties/maven.compiler.*
```

## Implementation

- `ChangeFilter` inner class in `PomChangeAnalyzer` — compiles glob patterns into `PathMatcher` instances
- Filter applied in `hasEffectiveChanges()` to each detected change path before marking a child as affected
- Filter applied to `changedProperties` at the parent level (feeds filtered resource checks)
- Reuses existing `parseList()` comma-separated pattern for configuration

## Tests

- 5 `ChangeFilter` unit tests (empty, exclude, glob, include-override, multiple patterns)
- 5 integration tests (property exclusion, git properties, include override, dependency exclusion, filtered resources, modules changes)
- 7 `ScalpelConfiguration` tests (defaults, custom values, empty override, known keys)

### DIFF
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
@@ -184,6 +184,44 @@ public final class ScalpelConfiguration {
      */
     public static final String MAX_RESOURCE_FILE_SIZE = PREFIX + "maxResourceFileSize";
 
+    /**
+     * System property {@code scalpel.excludeChanges}: comma-separated glob patterns over change
+     * paths that should be excluded from the effective model comparison. Excluded changes are
+     * treated as if they did not happen — a module is not marked as affected by them.
+     *
+     * <p>Change paths follow a normalized scheme:
+     * <ul>
+     *   <li>{@code properties/<name>} — e.g. {@code properties/build.timestamp}</li>
+     *   <li>{@code dependencies/<groupId>:<artifactId>} — e.g. {@code dependencies/com.foo:bar}</li>
+     *   <li>{@code plugins/<groupId>:<artifactId>}</li>
+     *   <li>{@code managedDependencies/<groupId>:<artifactId>}</li>
+     *   <li>{@code managedPlugins/<groupId>:<artifactId>}</li>
+     *   <li>{@code resources/<path>} — filtered resource files referencing changed properties</li>
+     * </ul>
+     *
+     * <p>Default: {@code properties/build.timestamp,properties/project.build.outputTimestamp}
+     * (volatile properties that change on every build and don't affect compilation).
+     *
+     * @see #INCLUDE_CHANGES
+     */
+    public static final String EXCLUDE_CHANGES = PREFIX + "excludeChanges";
+
+    /**
+     * System property {@code scalpel.includeChanges}: comma-separated glob patterns over change
+     * paths that force inclusion in the effective model comparison, even if the change would
+     * normally be ignored (e.g. because the property is inherited but not in the child's raw model).
+     *
+     * <p>Include patterns are applied after exclude patterns and override them: a change matching
+     * both an exclude and an include pattern is included.
+     *
+     * <p>Uses the same change path scheme as {@link #EXCLUDE_CHANGES}.
+     *
+     * <p>Default: none.
+     *
+     * @see #EXCLUDE_CHANGES
+     */
+    public static final String INCLUDE_CHANGES = PREFIX + "includeChanges";
+
     /** Mode value selecting {@code trim}: drop unaffected modules from the reactor. */
     public static final String MODE_TRIM = "trim";
 
@@ -205,6 +243,13 @@ public final class ScalpelConfiguration {
 
     /** Default value for {@link #REPORT_FILE}: {@code target/scalpel-report.json}. */
     private static final String DEFAULT_REPORT_FILE = "target/scalpel-report.json";
+
+    /**
+     * Default value for {@link #EXCLUDE_CHANGES}: volatile timestamp properties that change on
+     * every build and do not affect compilation output.
+     */
+    private static final String DEFAULT_EXCLUDE_CHANGES =
+            "properties/build.timestamp,properties/project.build.outputTimestamp";
 
     /** Default value for {@link #MAX_RESOURCE_FILE_SIZE}: ten mebibytes ({@code 10 * 1024 * 1024}). */
     public static final long DEFAULT_MAX_RESOURCE_FILE_SIZE = 10L * 1024 * 1024; // 10 MB
@@ -236,7 +281,9 @@ public final class ScalpelConfiguration {
             BUILD_ALL_IF_NO_CHANGES,
             IMPACTED_LOG,
             REPORT_FILE,
-            MAX_RESOURCE_FILE_SIZE)));
+            MAX_RESOURCE_FILE_SIZE,
+            EXCLUDE_CHANGES,
+            INCLUDE_CHANGES)));
 
     private final boolean enabled;
     private final String baseBranch;
@@ -265,6 +312,8 @@ public final class ScalpelConfiguration {
     private final boolean explain;
     private final String reportFile;
     private final long maxResourceFileSize;
+    private final List<String> excludeChanges;
+    private final List<String> includeChanges;
     private final List<String> warnings;
 
     private ScalpelConfiguration(
@@ -295,6 +344,8 @@ public final class ScalpelConfiguration {
             boolean explain,
             String reportFile,
             long maxResourceFileSize,
+            List<String> excludeChanges,
+            List<String> includeChanges,
             List<String> warnings) {
         this.enabled = enabled;
         this.baseBranch = baseBranch;
@@ -323,6 +374,8 @@ public final class ScalpelConfiguration {
         this.explain = explain;
         this.reportFile = reportFile;
         this.maxResourceFileSize = maxResourceFileSize;
+        this.excludeChanges = excludeChanges;
+        this.includeChanges = includeChanges;
         this.warnings = warnings;
     }
 
@@ -398,6 +451,9 @@ public final class ScalpelConfiguration {
             }
         }
 
+        List<String> excludeChanges = parseList(resolve(system, user, EXCLUDE_CHANGES, DEFAULT_EXCLUDE_CHANGES));
+        List<String> includeChanges = parseList(resolve(system, user, INCLUDE_CHANGES, null));
+
         List<String> warnings = detectUnknownKeys(system, user);
 
         return new ScalpelConfiguration(
@@ -428,6 +484,8 @@ public final class ScalpelConfiguration {
                 explain,
                 reportFile,
                 maxResourceFileSize,
+                excludeChanges,
+                includeChanges,
                 warnings);
     }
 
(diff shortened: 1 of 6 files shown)


extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
```diff
@@ -569,8 +660,12 @@ private void analyzeParentPomChange(
         Set<String> activeProfileIds = getActiveProfileIds(parentProject);
         parentSelfAffected = parentSelfAffected || analyzeProfileChanges(oldModel, newModel, activeProfileIds);
 
-        // Collect all changed properties
-        ctx.allChangedProperties.addAll(changedProperties);
+        // Collect all changed properties (filtered by change filter)
+        for (String prop : changedProperties) {
+            if (ctx.changeFilter.accepts("properties/" + prop)) {
+                ctx.allChangedProperties.add(prop);
```
gnodet, line 667: ⚠️ **Bug: filtered resource check still uses raw `changedProperties`**

This loop correctly filters `allChangedProperties`, but line 718 (below, same method scope) calls `hasFilteredResourcesWithChangedProperty(child, changedProperties, ctx)` with the **raw, unfiltered** `changedProperties` set. A property excluded here will still trigger a filtered-resource scan and can mark the child as affected — defeating the purpose of the exclusion.

Suggested fix: accumulate the filtered set here and use it below:

```suggestion
        // Collect all changed properties (filtered by change filter)
        Set<String> filteredChangedProperties = new LinkedHashSet<>();
        for (String prop : changedProperties) {
            if (ctx.changeFilter.accepts("properties/" + prop)) {
                ctx.allChangedProperties.add(prop);
                filteredChangedProperties.add(prop);
            }
        }
```
Then at the call site (~line 717) replace `changedProperties` with `filteredChangedProperties` in both the empty-check and the method call.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
```diff
@@ -184,6 +184,44 @@ public final class ScalpelConfiguration {
...
+     *
+     * <p>Change paths follow a normalized scheme:
+     * <ul>
+     *   <li>{@code properties/<name>} — e.g. {@code properties/build.timestamp}</li>
+     *   <li>{@code dependencies/<groupId>:<artifactId>} — e.g. {@code dependencies/com.foo:bar}</li>
+     *   <li>{@code plugins/<groupId>:<artifactId>}</li>
+     *   <li>{@code managedDependencies/<groupId>:<artifactId>}</li>
+     *   <li>{@code managedPlugins/<groupId>:<artifactId>}</li>
+     *   <li>{@code resources/<path>} — filtered resource files referencing changed properties</li>
+     * </ul>
```
gnodet, line 199: 🔴 **Javadoc documents a non-existent change path scheme**

The `resources/<path>` bullet is inaccurate — no code emits change paths with a `resources/` prefix. A user reading this and writing `-Dscalpel.excludeChanges=resources/**` gets silently no effect. Either implement it or remove this bullet.

extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzerTest.java
```diff
@@ -2703,6 +2722,335 @@ private List<MavenProject> createReactorWithBomImportCustomBom(Path root, String
...
+        Map<String, byte[]> oldPoms = new HashMap<>();
+        oldPoms.put("pom.xml", oldParentPom.getBytes(StandardCharsets.UTF_8));
+
+        // The property is excluded: even though filtered resources reference it,
+        // the property should not be in changedProperties
+        PomChangeAnalyzer.Result result =
+                analyzeChanges(changedPoms, oldPoms, projects, root, List.of("properties/build.timestamp"), List.of());
+        assertFalse(
+                result.getChangedProperties().contains("build.timestamp"),
+                "Excluded property should not appear in changedProperties");
```
gnodet, line 2976: 🔴 **Test validates the wrong observable — the bug slips through CI**

This only checks `result.getChangedProperties()`. It does not assert that the module is absent from `result.getAffectedProjects()`. Because `hasFilteredResourcesWithChangedProperty` is called with the raw set, a module with a filtered resource referencing the excluded property is still marked affected — and this test won't catch it.

Add:
```java
assertFalse(
    result.getAffectedProjects().contains(projects.get(2)),
    "Excluded property must not trigger filtered-resource scan");
```

gnodet (commented):
Two real issues in this PR.

**1. Filtered resource check bypasses the exclude filter (production bug)**

The change filter is correctly applied when collecting `allChangedProperties` (lines 664–668 in `analyzeParentPomChange`), but `hasFilteredResourcesWithChangedProperty` at line 718 receives the **raw, unfiltered** `changedProperties` set. This means: if a user excludes `properties/build.timestamp` but a child module has a filtered resource referencing `${build.timestamp}`, the child **will still be marked as affected** — the exclusion has no effect on the filtered-resource code path.

The PR description says _"Filter applied to changedProperties at the parent level (feeds filtered resource checks)"_ — this claim is wrong. `allChangedProperties` feeds the `Result` object, not the filtered-resource scan.

The test `analyzeChanges_excludePropertyAlsoExcludesFromFilteredResources` validates only `result.getChangedProperties()`, not `result.getAffectedProjects()`. The assert checks the wrong thing and the bug slips through CI.

Fix — filter before calling `hasFilteredResourcesWithChangedProperty`:
```java
// Compute properties that pass the change filter
Set<String> filteredChangedProperties = new LinkedHashSet<>();
for (String prop : changedProperties) {
    if (ctx.changeFilter.accepts("properties/" + prop)) {
        ctx.allChangedProperties.add(prop);
        filteredChangedProperties.add(prop);
    }
}
```
Then pass `filteredChangedProperties` (instead of `changedProperties`) to `hasFilteredResourcesWithChangedProperty`, and update the empty-check guard to match.

Add a test assertion: `assertFalse(result.getAffectedProjects().contains(moduleB), "excluded property must not trigger filtered-resource scan")`.

**2. `resources/<path>` documented but not implemented**

The Javadoc for `EXCLUDE_CHANGES` lists `resources/<path>` as a valid change path scheme. But no code generates change paths with a `resources/` prefix — `hasFilteredResourcesWithChangedProperty` doesn't emit them. A user writing `-Dscalpel.excludeChanges=resources/**` gets silently no effect. Either implement it or remove this bullet from the Javadoc.

---

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
gnodet: Fixed in f751984. Introduced `filteredChangedProperties` local set and wired it into both the empty-check guard and the `hasFilteredResourcesWithChangedProperty` call.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
gnodet: Fixed in f751984. Removed the `resources/<path>` bullet from the Javadoc.

extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzerTest.java
gnodet: Fixed in f751984. Added `Resource` with `filtering=true` setup so the code path is exercised, plus the `assertFalse(affectedProjects.contains(moduleB))` assertion.

PR #176, opened 2026-09-07
Fix #124: centralise path filtering into a single PathFilters type

### LINKED ISSUE #124: [arch] includePaths filtering re-derived at four sites with subtly different rules (correctness risk wearing a refactoring costume)
## Problem

`includePaths` filtering is independently re-derived and re-applied at four sites with subtly different rules: the affected-set stage; the trim-mode build-set stage, where upstream prerequisites survive; the skip-tests stage, where they do NOT survive; and the report's downstream section. A comment at the first site explicitly claims the matchers are "compiled once here and reused later" while the report site recompiles from config, contradicting it.

Any future change to include semantics must be made in four places and kept consistent by hand, and the divergence between the trim and skip-tests sites suggests that has already failed once. Glob compilation is also scattered, with matchers compiled inline in loops rather than via the existing helper, and changed-file paths reparsed once per pattern.

## Proposal

A single `PathFilters` type owning compilation and evaluation of exclude/include/disable-trigger/full-build-trigger patterns, constructed once and passed down. The one intentional semantic difference (upstream prerequisites surviving in trim mode) expressed explicitly and documented rather than left implicit. Changed files converted to `Path` once. Behavior preserved, proven by the existing suite plus an include-paths IT (see the IT-coverage issue).

gnodet:
## Summary

Introduce `PathFilters`, a single type that compiles all glob patterns (`excludePaths`, `includePaths`, `disableTriggers`, `fullBuildTriggers`) once and provides evaluation methods for all four sites that previously re-derived them independently.

Fixes #124.

## Changes

- **New class `PathFilters`**: compiles all glob patterns once in the constructor and exposes:
  - `filterExcludedPaths()` — filters changed files against exclude patterns
  - `matchesDisableTrigger()` — checks if any changed file triggers disabling
  - `findFullBuildTrigger()` — finds the first file triggering a full build
  - `matchesIncludePaths()` — strict module-level include check (skip-tests mode)
  - `filterBuildSet()` — trim-mode filtering that preserves upstream prerequisites
  - `hasIncludeFilters()` — whether includePaths are configured

- **Refactored `ScalpelLifecycleParticipant`**:
  - Creates `PathFilters` once early in `afterProjectsRead` and threads it to all call sites
  - Removed 5 private methods that are now in `PathFilters`
  - Removed unused imports and constant
  - `addDownstreamModules` no longer recompiles matchers from config
  - Changed files are converted to `Path` once per evaluation rather than once per pattern

- **Documented semantic difference**:
  - **Trim mode** (`filterBuildSet`): upstream prerequisites survive the include filter
  - **Skip-tests mode** (`matchesIncludePaths`): no exception for upstream — tests are skipped

- **26 new unit tests** for `PathFilters`

## Stats

- `ScalpelLifecycleParticipant.java`: 34 insertions, 106 deletions (net -72 lines)
- `PathFilters.java`: 236 lines (new)
- `PathFiltersTest.java`: 360 lines (new)
- All 216 existing + new tests pass

## Summary by CodeRabbit

- **Improvements**
  - Improved consistency when applying include and exclude path filters across build modes.
  - Preserved correct handling of disable triggers and full-build triggers.
  - Refined glob pattern matching, including patterns with and without directory separators.
  - Ensured trim mode retains affected and upstream modules when appropriate, while skip-tests mode applies stricter include-path filtering.

- **Tests**
  - Added comprehensive coverage for path filtering, trigger detection, glob normalization, and build-set behavior.

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PathFilters.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PathFilters.java
@@ -1,0 +1,220 @@
+/*
+ * Copyright (c) Maveniverse Org.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-v20.html
+ */
+package eu.maveniverse.maven.scalpel.extension3.internal;
+
+import static java.util.Objects.requireNonNull;
+
+import eu.maveniverse.maven.scalpel.core.ScalpelConfiguration;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import org.apache.maven.project.MavenProject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Centralises all glob-based path filtering for Scalpel: exclude paths, include paths,
+ * disable triggers, and full-build triggers. All glob patterns are compiled once in the
+ * constructor and reused across every evaluation site, replacing the previous approach of
+ * recompiling matchers independently at each call site.
+ *
+ * <h3>Trim mode vs skip-tests mode</h3>
+ * <p>There is one <em>intentional</em> semantic difference between how include-path filtering
+ * is applied in <strong>trim mode</strong> and <strong>skip-tests mode</strong>:</p>
+ * <ul>
+ *   <li><strong>Trim mode</strong> ({@link #filterBuildSet}): upstream build prerequisites
+ *       survive the include filter even when their path falls outside {@code includePaths},
+ *       because removing them would break the build (missing compile-time dependencies).
+ *       Downstream modules outside the scope <em>are</em> dropped.</li>
+ *   <li><strong>Skip-tests mode</strong> ({@link #matchesIncludePaths}): every module outside
+ *       {@code includePaths} has its tests skipped, with <em>no</em> exception for upstream
+ *       prerequisites. The upstream module is still built (skip-tests mode never removes
+ *       modules from the reactor), but its tests are not run.</li>
+ * </ul>
+ */
+class PathFilters {
+
+    private static final String GLOB_PREFIX = "glob:";
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    private final List<PathMatcher> excludeMatchers;
+    private final List<PathMatcher> includeMatchers;
+    private final List<String> disableTriggerPatterns;
+    private final List<PathMatcher> disableTriggerMatchers;
+    private final List<String> fullBuildTriggerPatterns;
+    private final List<PathMatcher> fullBuildTriggerMatchers;
+
+    PathFilters(ScalpelConfiguration config) {
+        requireNonNull(config, "config");
+        this.excludeMatchers = compileGlobMatchers(config.getExcludePaths());
+        this.includeMatchers = compileGlobMatchers(config.getIncludePaths());
+        this.disableTriggerPatterns = List.copyOf(config.getDisableTriggers());
+        this.disableTriggerMatchers = compileGlobMatchers(config.getDisableTriggers());
+        this.fullBuildTriggerPatterns = List.copyOf(config.getFullBuildTriggers());
+        this.fullBuildTriggerMatchers = compileGlobMatchers(config.getFullBuildTriggers());
+    }
+
+    /**
+     * Filters out changed files matching any {@code excludePaths} glob pattern.
+     *
+     * @param changedFiles the original set of changed file paths (forward-slash separated)
+     * @return a new set with excluded files removed; the original set when no excludes are
+     *     configured
+     */
+    Set<String> filterExcludedPaths(Set<String> changedFiles) {
+        if (excludeMatchers.isEmpty()) {
+            return changedFiles;
+        }
+        Set<String> filtered = new LinkedHashSet<>();
+        for (String file : changedFiles) {
+            Path p = Path.of(file);
+            boolean excluded = false;
+            for (PathMatcher matcher : excludeMatchers) {
+                if (matcher.matches(p)) {
+                    excluded = true;
+                    break;
+                }
+            }
+            if (!excluded) {
+                filtered.add(file);
+            }
+        }
+        int excludedCount = changedFiles.size() - filtered.size();
+        if (excludedCount > 0) {
+            logger.info("Scalpel: {} files excluded by path filters", excludedCount);
+        }
+        return filtered;
+    }
+
+    /**
+     * Returns {@code true} if any changed file matches a {@code disableTriggers} glob pattern.
+     */
+    boolean matchesDisableTrigger(Set<String> changedFiles) {
+        if (disableTriggerMatchers.isEmpty()) {
+            return false;
+        }
+        for (int i = 0; i < disableTriggerMatchers.size(); i++) {
+            PathMatcher matcher = disableTriggerMatchers.get(i);
+            String pattern = disableTriggerPatterns.get(i);
+            for (String changedFile : changedFiles) {
+                Path p = Path.of(changedFile);
+                if (matcher.matches(p)) {
+                    logger.info(
+                            "Scalpel: Disabled due to change in {} (matches disable trigger {})", changedFile, pattern);
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Returns the first changed file matching a {@code fullBuildTriggers} glob pattern, or
+     * {@code null} if none match.
+     */
+    String findFullBuildTrigger(Set<String> changedFiles) {
+        if (fullBuildTriggerMatchers.isEmpty()) {
+            return null;
+        }
+        for (int i = 0; i < fullBuildTriggerMatchers.size(); i++) {
+            PathMatcher matcher = fullBuildTriggerMatchers.get(i);
+            String pattern = fullBuildTriggerPatterns.get(i);
+            for (String changedFile : changedFiles) {
+                Path p = Path.of(changedFile);
+                if (matcher.matches(p)) {
+                    logger.info("Scalpel: Full build triggered by change to {} (matches {})", changedFile, pattern);
+                    return changedFile;
+                }
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Returns whether any {@code includePaths} patterns are configured.
+     */
+    boolean hasIncludeFilters() {
+        return !includeMatchers.isEmpty();
+    }
+
+    /**
+     * Returns whether the given project's module path matches any {@code includePaths} glob
+     * pattern. This is the <strong>strict</strong> check used in skip-tests mode: no exception
+     * is made for upstream prerequisites.
+     *
+     * @param project the Maven project to test
+     * @param reactorRoot the reactor root directory
+     * @return {@code true} if the module path matches at least one include pattern
+     */
+    boolean matchesIncludePaths(MavenProject project, Path reactorRoot) {
+        String relPath = relativePath(reactorRoot, project);
+        Path modulePath = Path.of(relPath);
+        for (PathMatcher matcher : includeMatchers) {
+            if (matcher.matches(modulePath) || matcher.matches(modulePath.resolve("pom.xml"))) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Filters a build set for <strong>trim mode</strong>: removes downstream modules outside
+     * {@code includePaths} scope while keeping directly/transitively affected modules and
+     * upstream build prerequisites. Upstream prerequisites survive even when their path falls
+     * outside the include scope, because removing them would break compilation.
+     *
+     * @param buildSet the full build set from {@code ReactorTrimmer}
+     * @param affected the set of directly and transitively affected modules
+     * @param upstreamOnly the set of modules classified as upstream-only by the trimmer
+     * @param reactorRoot the reactor root directory
+     * @return the filtered build set (a new mutable list); the original list when no include
+     *     filters are configured
+     */
+    List<MavenProject> filterBuildSet(
+            List<MavenProject> buildSet, Set<MavenProject> affected, Set<MavenProject> upstreamOnly, Path reactorRoot) {
+        if (includeMatchers.isEmpty()) {
+            return buildSet;
+        }
+        List<MavenProject> filtered = new ArrayList<>(buildSet);
+        filtered.removeIf(project -> !affected.contains(project)
+                && !upstreamOnly.contains(project)
+                && !matchesIncludePaths(project, reactorRoot));
+        return filtered;
+    }
+
+    // ---- internal helpers ----
+
+    private static List<PathMatcher> compileGlobMatchers(List<String> patterns) {
+        if (patterns.isEmpty()) {
+            return List.of();
+        }
+        List<PathMatcher> matchers = new ArrayList<>(patterns.size());
+        for (String pattern : patterns) {
+            matchers.add(FileSystems.getDefault()
+                    .getPathMatcher(GLOB_PREFIX + ChangedFileClassifier.normalizeGlobPattern(pattern)));
+        }
+        return matchers;
+    }
+
+    /**
+     * Computes the relative path from the reactor root to the project's base directory.
+     * The root must already be normalized via {@code Path.toAbsolutePath().normalize()} — this
+     * method does NOT re-normalize the root on every call (#113).
+     */
+    private static String relativePath(Path normalizedRoot, MavenProject project) {
+        return normalizedRoot
+                .relativize(project.getBasedir().toPath().toAbsolutePath().normalize())
+                .toString()
+                .replace('\\', '/');
+    }
+}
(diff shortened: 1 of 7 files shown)


extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PathFilters.java
```diff
@@ -0,0 +1,236 @@
...
+        Set<String> filtered = new LinkedHashSet<>();
+        int idx = 0;
+        for (String file : changedFiles) {
+            Path p = paths.get(idx++);
+            boolean excluded = false;
+            for (PathMatcher matcher : excludeMatchers) {
+                if (matcher.matches(p)) {
+                    excluded = true;
+                    break;
+                }
```
gnodet, line 86: ⚠️ **Fragile parallel-index coupling — silently wrong for non-ordered Sets**

`toPaths(changedFiles)` iterates `changedFiles` to build the list; this loop iterates `changedFiles` again with `idx++` to fetch the pre-built `Path`. This relies on both iterations visiting elements in the same order. `LinkedHashSet` guarantees this, but the method signature accepts any `Set<String>`, and there is no guard against unordered inputs.

Simpler and correct:
```suggestion
        Set<String> filtered = new LinkedHashSet<>();
        for (String file : changedFiles) {
            Path p = Path.of(file);
            boolean excluded = false;
            for (PathMatcher matcher : excludeMatchers) {
                if (matcher.matches(p)) {
                    excluded = true;
                    break;
                }
            }
            if (!excluded) {
                filtered.add(file);
            }
        }
```
Same applies to `matchesDisableTrigger` and `findFullBuildTrigger` — replace `toPaths()` + `idx++` with `Path.of(file)` inline.

gnodet (commented):
One structural issue worth fixing before merge.

**Parallel `toPaths()` / `idx++` pattern is a latent correctness hazard**

`filterExcludedPaths`, `matchesDisableTrigger`, and `findFullBuildTrigger` all call `toPaths(changedFiles)` to build a `List<Path>` and then iterate `changedFiles` again in lock-step using `idx++`. This assumes the two iterations visit elements in the same order. That holds for `LinkedHashSet` (which is what all current callers use), but breaks silently for any unordered `Set`. No compiler or runtime guard catches the mismatch — the matcher just compares the wrong `Path` to the wrong file string.

The simpler and obviously-correct approach is to convert inline:
```java
for (String file : changedFiles) {
    Path p = Path.of(file);
    // ... match p against matchers
}
```
This is the same cost (one `Path.of()` per file per call, identical to what `toPaths()` does) and eliminates the fragile parallel-index coupling. The `toPaths()` helper and the `paths.get(idx++)` pattern can be removed entirely.

---

gnodet:
Addressed the `toPaths()` / `idx++` parallel-index pattern: converted to inline `Path.of(file)` in each loop, removed the `toPaths()` helper entirely. Fixed in 8041793.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PathFilters.java
```diff
@@ -0,0 +1,217 @@
...
+        }
+        List<PathMatcher> matchers = new ArrayList<>(patterns.size());
+        for (String pattern : patterns) {
+            matchers.add(FileSystems.getDefault()
+                    .getPathMatcher(GLOB_PREFIX + ScalpelLifecycleParticipant.normalizeGlobPattern(pattern)));
+        }
+        return matchers;
+    }
+
+    static String relativePath(Path reactorRoot, MavenProject project) {
```
gnodet: 🔸 **Duplication / visibility nit** — `relativePath` is declared `static` (package-private) but is only called from `matchesIncludePaths` within this class. `ScalpelLifecycleParticipant` still has its own `private static relativePath` at line 1846 with an identical implementation. The PR removed `matchesIncludePaths`, `filterExcludedPaths`, `findFullBuildTrigger`, `matchesDisableTrigger`, and `compileGlobMatchers` from SLP, but `relativePath` was left behind in both places.

Minimum fix — make it `private static` since it has no external callers:

```suggestion
    private static String relativePath(Path reactorRoot, MavenProject project) {
```

Optional follow-up: update SLP to delegate to `PathFilters.relativePath(...)` and remove the duplicate. Not blocking if it's left as a separate cleanup.

gnodet (commented):
**Re-review of PR 176 after rebase (HEAD: 8041793)**

The branch was rebased; the diff between last-reviewed SHA (`1a45aea`) and current HEAD only touches `ScalpelReportTest.java`, which is not part of this PR's change set. The PathFilters content is identical to what was reviewed previously.

**Previous finding: `toPaths()/idx++` parallel-index coupling** ✅ Addressed

Confirmed: the `toPaths()` helper and all `idx++` usages are gone. `filterExcludedPaths`, `matchesDisableTrigger`, and `findFullBuildTrigger` now all use `Path.of(file)` inline per iteration. The fix is correct and matches the suggested approach.

**New finding: `relativePath` duplication**

See inline comment. `PathFilters.relativePath` is declared `static` (package-private) but is only called from within `PathFilters` itself. `ScalpelLifecycleParticipant` still contains its own `private static relativePath` method at line 1846. The PR removed 5 private methods from SLP but left its `relativePath` in place while adding a second copy to `PathFilters`. This is the only remaining issue.

**ast-grep findings (ScalpelLifecycleParticipant.java):** Five `broad-exception-catch` hits at lines 87, 269, 664, 1575, 1598 and one `empty-catch-block` at line 269 — all pre-existing, not introduced by this PR.

paoloantinori:
Reviewed this against our own (now-closed) implementation of the same consolidation, and your version holds up: the excluded-count line and both trigger lines keep their exact `Scalpel: `​prefixed wording, which our first pass had silently dropped (external automation greps those, so it is worth an assertion if you add a PathFiltersTest case for it).

Two observations from ours that may transfer:

1. **triggerFile selection order.** If your trigger search is file-major (for each file, for each pattern) rather than pattern-major (the old code's shape: for each pattern, for each file), then with multiple patterns and multiple changed files the reported `triggerFile` can differ from before, and it lands in the report's `triggerFile` field and the shadow status reason. Harmless, but worth one line in the PR body if it applies.

2. **normalizeGlobPattern placement.** We found `ScalpelLifecycleParticipant.normalizeGlobPattern` had no remaining in-class callers after the migration, so it moved naturally into PathFilters (two tests requalified). Yours may already do this.

Happy to close ours as duplicate of this one; it was second in time.

gnodet:
Addressed: rebased on main's collaborator refactoring (#174). The `toPaths()/idx++` pattern was replaced with inline `Path.of(file)` in the original commit. The rebase integrates `PathFilters` alongside `ChangedFileClassifier`: pattern normalization stays in `ChangedFileClassifier`, while `PathFilters` owns the compile-once responsibility passed through to `ReportAssembler` and `SkipTestsApplier`.

extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmerTest.java
```diff
@@ -197,170 +197,6 @@ void computeBuildSet_buildOrderPreserved() {
         assertEquals(List.of(projectA, projectC), result.getBuildSet());
```
gnodet, line 197: 🐛 **Accidental deletion during rebase** — These 5 tests (`transitiveDownstreamViaBFS`, `transitiveUpstreamViaBFS`, `testOnlyDoesNotPropagateTransitively`, `diamondGraphNoDuplicates`, `testOnlyAndNonTestOnlyConvergingPaths`) were added by PR #173 to cover the BFS traversal logic in `ReactorTrimmer`. They are unrelated to the `PathFilters` refactoring and were lost during the rebase conflict resolution. Please restore them.

gnodet (commented):
**Re-review after rebase onto main (HEAD: 3fd17a3)**

The `PathFilters` extraction itself looks clean — pattern compilation once in the constructor, five evaluation methods replacing scattered recompilation at each call site, well-documented trim vs skip-tests semantic difference. The test coverage is thorough.

Two things to fix:

### 🐛 Accidental deletion of ReactorTrimmer BFS tests

The rebase onto main (which now contains PR #173) produced a conflict in `ReactorTrimmerTest.java`. The resolution **deleted** 5 BFS traversal tests that are unrelated to `PathFilters`:

- `computeBuildSet_transitiveDownstreamViaBFS`
- `computeBuildSet_transitiveUpstreamViaBFS`
- `computeBuildSet_testOnlyDoesNotPropagateTransitively`
- `computeBuildSet_diamondGraphNoDuplicates`
- `computeBuildSet_testOnlyAndNonTestOnlyConvergingPaths`

These tests were added by PR #173 and exercise `ReactorTrimmer.computeBuildSet()` — they have nothing to do with path filtering. They need to be restored.

### 🧹 Dead code left on `ChangedFileClassifier`

The old path-filtering methods on `ChangedFileClassifier` are now dead code:
- `matchesDisableTrigger(Set<String>, ScalpelConfiguration)` — replaced by `PathFilters.matchesDisableTrigger()`
- `filterExcludedPaths(Set<String>, ScalpelConfiguration)` — replaced by `PathFilters.filterExcludedPaths()`
- `findFullBuildTrigger(Set<String>, ScalpelConfiguration)` — replaced by `PathFilters.findFullBuildTrigger()`
- `matchesIncludePaths(MavenProject, List<PathMatcher>, Path)` — replaced by `PathFilters.matchesIncludePaths()`
- `compileGlobMatchers(List<String>)` — now only called from `PathFilters` internal helper (which has its own copy)

Only `classifyChanges()`, `relativePath()`, and `normalizeGlobPattern()` are still called. The rest should be removed to complete the extraction — leaving them invites confusion about which code path is authoritative.

(Minor: `PathFiltersTest.normalizeGlobPattern_*` tests call `ScalpelLifecycleParticipant.normalizeGlobPattern()` which is just a delegate to `ChangedFileClassifier.normalizeGlobPattern()`. Consider testing the source directly, or better, make `normalizeGlobPattern` package-private on `PathFilters` since that's where it's actually used.)

gnodet (commented):
**Re-review after merge commit `8a1dbb7` (HEAD merges `origin/main` into the PR branch)**

The merge integrates the monolithic refactoring from main (where `ChangedFileClassifier`, `ReportAssembler`, `SkipTestsApplier`, `TransitiveImpactAnalyzer` etc. were folded back into `ScalpelLifecycleParticipant`). `PathFilters` itself is essentially unchanged — only two lines touched: `ChangedFileClassifier.normalizeGlobPattern` → `ScalpelLifecycleParticipant.normalizeGlobPattern`, and `relativePath` now re-normalizes the root on every call. Both are correct given the new codebase shape.

### 🐛 ReactorTrimmer BFS tests still deleted (unfixed from previous review)

The merge resolution **still** drops 5 BFS traversal tests from `ReactorTrimmerTest.java`:

- `computeBuildSet_transitiveDownstreamViaBFS`
- `computeBuildSet_transitiveUpstreamViaBFS`
- `computeBuildSet_testOnlyDoesNotPropagateTransitively`
- `computeBuildSet_diamondGraphNoDuplicates`
- `computeBuildSet_testOnlyAndNonTestOnlyConvergingPaths`

These tests exist on `main` (added by PR #173) and exercise `ReactorTrimmer.computeBuildSet()` — they have nothing to do with path filtering. The merge dropped them (likely a conflict resolution error). The PR branch has 7 tests vs 12 on main. **This must be fixed before merge.**

### ✅ Paolo's observations — both non-issues

1. **Trigger file selection order**: `matchesDisableTrigger` and `findFullBuildTrigger` are pattern-major (outer loop over matchers, inner loop over files) — same iteration shape as the old `ChangedFileClassifier` code. The reported `triggerFile` is deterministic and matches previous behavior.

2. **`normalizeGlobPattern` placement**: After the monolithic refactoring on main, `ChangedFileClassifier` no longer exists — `normalizeGlobPattern` lives on `ScalpelLifecycleParticipant` (line 715). `PathFilters.compileGlobMatchers` correctly references `ScalpelLifecycleParticipant.normalizeGlobPattern()`. No remaining in-class callers issue.

### 🧹 Minor: `relativePath` duplication persists

`PathFilters.relativePath` (line 209, calls `toAbsolutePath().normalize()` on every invocation) vs `ScalpelLifecycleParticipant.relativePath` (line 1848, expects pre-normalized root). Both are correct but have subtly different semantics. Not a blocker — the SLP version is the hot-path one (called in loops), and it correctly receives `normalizedRoot`. The `PathFilters` version is called only via `matchesIncludePaths` which receives the raw `reactorRoot`, so the re-normalization is needed there. Just worth documenting in a comment on one of them.

### Bottom line

The only blocking issue is the missing BFS tests — everything else is clean. Once those 5 tests are restored, LGTM.

gnodet:
Thanks @paoloantinori for the thorough cross-check against your implementation!

On your two observations:

1. **Trigger file selection order** — confirmed pattern-major in `PathFilters.findFullBuildTrigger` and `matchesDisableTrigger` (outer loop over compiled matchers, inner loop over changed files), so the reported `triggerFile` is deterministic and matches the old `ChangedFileClassifier` behavior. No change needed.

2. **`normalizeGlobPattern` placement** — after the monolithic refactoring landed on main (PR #174 and follow-ups), `ChangedFileClassifier` was absorbed into `ScalpelLifecycleParticipant`. `normalizeGlobPattern` now lives on SLP at line 715, and `PathFilters.compileGlobMatchers` references it from there. The placement is clean.

Happy to take the duplicate closure — this one just needs the 5 BFS tests restored from the merge conflict resolution and it should be good to go.

gnodet:
Thanks @paoloantinori for the thorough cross-check. Addressed both observations and the outstanding review items in 354abd8:

1. **Trigger iteration order**: confirmed the current code uses **pattern-major** iteration (for each pattern, for each file) — same shape as the old code. So `triggerFile` selection order is preserved. No semantic change there.

2. **`normalizeGlobPattern` placement**: pattern normalization stays on `ChangedFileClassifier` (where it was added by #174's collaborator extraction). `PathFilters` delegates to it via `ChangedFileClassifier.normalizeGlobPattern()`. The `SLP.normalizeGlobPattern` one-liner delegate is kept for now to avoid churn in existing tests; can be cleaned up separately.

Also in this commit:
- **Restored 5 BFS traversal tests** that were accidentally dropped during the rebase conflict resolution (`transitiveDownstreamViaBFS`, `transitiveUpstreamViaBFS`, `testOnlyDoesNotPropagateTransitively`, `diamondGraphNoDuplicates`, `testOnlyAndNonTestOnlyConvergingPaths`)
- **Removed dead code** from `ChangedFileClassifier`: the 5 methods now replaced by `PathFilters` (`matchesDisableTrigger`, `filterExcludedPaths`, `findFullBuildTrigger`, `matchesIncludePaths`, `compileGlobMatchers`) and their tests
- **Made `PathFilters.relativePath` private** — only called internally; external callers use `ChangedFileClassifier.relativePath()`

All 216+ tests pass.

gnodet (commented):
**Re-review after fix commit `354abd8` (HEAD: 354abd8)**

This commit addresses both findings from the previous review. Verified the delta between `3fd17a3` (last-reviewed SHA) and `354abd8`.

### ✅ ReactorTrimmer BFS tests restored

All 5 BFS traversal tests are back in `ReactorTrimmerTest.java`:

- `computeBuildSet_transitiveDownstreamViaBFS`
- `computeBuildSet_transitiveUpstreamViaBFS`
- `computeBuildSet_testOnlyDoesNotPropagateTransitively`
- `computeBuildSet_diamondGraphNoDuplicates`
- `computeBuildSet_testOnlyAndNonTestOnlyConvergingPaths`

The PR branch now has 12 tests, matching `origin/main`.

### ✅ Dead code removed from `ChangedFileClassifier`

`ChangedFileClassifier` went from 178 → 80 lines. The 5 dead methods (`matchesDisableTrigger`, `filterExcludedPaths`, `findFullBuildTrigger`, `matchesIncludePaths`, `compileGlobMatchers`) plus the `GLOB_PREFIX` constant and associated imports are all removed. What remains:

- `classifyChanges()` — actively used
- `normalizeGlobPattern()` — called by `PathFilters.compileGlobMatchers`
- `relativePath()` — called by SLP at line 451
- `ClassificationResult` inner class

Corresponding tests in `ChangedFileClassifierTest` were also cleaned up (removed tests for methods that no longer exist on this class; equivalent coverage now lives in `PathFiltersTest`).

### ✅ `PathFilters.relativePath` now `private`

The previous review noted `relativePath` duplication between `PathFilters` and `ChangedFileClassifier`. Both copies have identical semantics (expects pre-normalized root, re-normalizes project basedir). `PathFilters.relativePath` is now correctly scoped `private` — it's an internal helper. `ChangedFileClassifier.relativePath` remains package-private for SLP's use. Duplication is minor and justified by the encapsulation boundary.

### ⚠️ Merge conflict with `main`

GitHub reports `CONFLICTING` mergeable state. `main` has one new commit since the branch's merge-base (`fe0ae1b` — #177 periodic full-build verification), which touches `ScalpelLifecycleParticipant.java` and `ReportAssembler.java` — both also modified by this PR. A rebase is needed before merge.

### Bottom line

All previous review findings are addressed. The only remaining action is rebasing onto current `main` to resolve the merge conflict. Code LGTM.

gnodet (commented):
**Re-review after rebase (HEAD: 1df9671)**

The rebase onto current `main` resolves the merge conflict noted in the previous review. The PR is now a clean single squashed commit.

No code changes from the previously-reviewed version — only the rebase conflict resolution (integration with #177's changes to `ChangesetContext` and `ReportAssembler`).

All previous findings (dead-code cleanup in `ChangedFileClassifier`, BFS test restoration, `relativePath` scoping) were addressed in the prior iteration. CI unit tests passing; integration tests in progress.

LGTM ✅

_Hermès on behalf of @gnodet_

PR #177, opened 2026-09-07
Fix #101: periodic full-build verification and a stable decisionId

### LINKED ISSUE #101: [enh] Periodic full-build verification mode and a stable decisionId
## Problem

There is no supported way to answer "was Scalpel right last week?" before a team gates a release branch on trim mode. No decision fingerprint exists to correlate a later failure back to a Scalpel decision.

## Proposal

- `-Dscalpel.verifyFullBuild=true` for a scheduled job: run the full build, compute the would-be decision, and exit non-zero if any module Scalpel would have skipped actually fails, reporting the modules and why they were judged skippable.
- A stable `decisionId` (hash of merge base, head, resolved config and resulting build set) emitted in every mode, so a decision is quotable in a bug report and correlatable with a post-merge failure.
- A ready-to-use scheduled GitHub Actions job documented.

Would build naturally on a shadow mode (shadow plus non-zero exit code). Happy to help.

paoloantinori:
Implements the issue's three asks:

**`-Dscalpel.verifyFullBuild=true`** runs the full build under the shadow observation (reactor untouched, all tests run, whatever the configured mode was; trim and skip-tests are overridden with a warning) and fails the run when a module Scalpel would have skipped fails, naming each module, the reason it was judged skippable, and the decision id. The failure is added to the session result, so the exit code is driven by the Scalpel verdict and survives `-fn`-style invocations. This is the scheduled-job answer to "was Scalpel right last week?".

**`decisionId`**: every run in every mode emits a stable SHA-256 identity of the decision (merge-base commit, head commit, the resolved decision-shaping configuration fingerprint, and the sorted build-set paths) in the JSON report (optional additive field, schema and drift guard updated), in the shadow document and in each history line; the report and the shadow document of one run quote the same id. `ChangeDetectionResult` now carries the merge-base and head commit ids; the fingerprint covers exactly the set-shaping properties (sorted, order-insensitive) and excludes output-shaping ones like reportFile or mode, so a verify run and the later trim run on the same changeset share the id.

**A ready-to-use scheduled job** is documented in CI Recipes (nightly `scalpel-verify` workflow with artifact upload of the shadow documents on failure).

TDD: the configuration flag and fingerprint tests and the decisionId determinism/order-insensitivity/null-tolerance tests were written first and demonstrated failing at compile level on unmodified main.

Gates run before pushing (simplify four angles, review: id-stability, verify escape hatches, addException bytecode semantics, fingerprint completeness). Two real defects found and fixed by the gates: trim/report hashed the root aggregator as the empty string while shadow/verify hashed it as ".", so the same decision produced different ids across modes (every mode now collects through one `.\`-normalizing key helper); the verify-failure IT passed on surefire's failure alone and now asserts the monitor's own verdict line. The fingerprint also gained `maxResourceFileSize` (a resource over the cap is skipped, changing the affected set).

* a verify run overrides the reactor-modifying modes with a WARN in the log
* the report schema gains the optional `decisionId` property (no version bump per the additive policy; golden file and bidirectional drift guard updated)
* shadow/verify status-document overwrites now also cover the bail-outs of verify runs

Full battery green: core 203, extension3 190 unit tests, all 38 integration tests (two new: verify-success, verify-failure).

## Summary by CodeRabbit

- **New Features**
  - Added full-build verification mode to detect failures in modules that would otherwise be skipped.
  - Reports now include stable decision IDs for correlating build results and failures.
  - Verification results identify skipped-module reasons and false-negative failures.
  - Added strict configuration support for enabling full-build verification.

- **Documentation**
  - Added configuration, report-format, and scheduled verification guidance.

- **Tests**
  - Added coverage for successful and failing verification runs, report output, and decision-ID consistency.

### DIFF
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
@@ -79,6 +79,14 @@ public final class ScalpelConfiguration {
     public static final String EXPLAIN = PREFIX + "explain";
 
     /**
+     * System property {@code scalpel.verifyFullBuild}: run the full build, compute the would-be
+     * decision, and fail the build if any module Scalpel would have skipped fails, naming the
+     * modules and why they were judged skippable (#101). Forces the shadow observation instead
+     * of reactor-modifying modes. Default: {@code false}.
+     */
+    public static final String VERIFY_FULL_BUILD = PREFIX + "verifyFullBuild";
+
+    /**
      * System property {@code scalpel.disableOnBranch}: comma-separated regex patterns; Scalpel is
      * disabled when the current branch matches any of them. Default: none.
      */
@@ -219,6 +227,7 @@ public final class ScalpelConfiguration {
             FAIL_SAFE,
             MODE,
             EXPLAIN,
+            VERIFY_FULL_BUILD,
             DISABLE_ON_BRANCH,
             DISABLE_ON_BASE_BRANCH,
             EXCLUDE_PATHS,
@@ -263,6 +272,7 @@ public final class ScalpelConfiguration {
     private final boolean failSafe;
     private final String mode;
     private final boolean explain;
+    private final boolean verifyFullBuild;
     private final String reportFile;
     private final long maxResourceFileSize;
     private final List<String> warnings;
@@ -293,6 +303,7 @@ public final class ScalpelConfiguration {
             boolean failSafe,
             String mode,
             boolean explain,
+            boolean verifyFullBuild,
             String reportFile,
             long maxResourceFileSize,
             List<String> warnings) {
@@ -321,6 +332,7 @@ public final class ScalpelConfiguration {
         this.failSafe = failSafe;
         this.mode = mode;
         this.explain = explain;
+        this.verifyFullBuild = verifyFullBuild;
         this.reportFile = reportFile;
         this.maxResourceFileSize = maxResourceFileSize;
         this.warnings = warnings;
@@ -382,6 +394,8 @@ public final class ScalpelConfiguration {
                     + ", " + MODE_SKIP_TESTS + ", " + MODE_REPORT + ", " + MODE_SHADOW);
         }
         boolean explain = parseStrictBoolean(EXPLAIN, resolve(system, user, EXPLAIN, "false"));
+        boolean verifyFullBuild =
+                parseStrictBoolean(VERIFY_FULL_BUILD, resolve(system, user, VERIFY_FULL_BUILD, "false"));
         String reportFile = resolve(system, user, REPORT_FILE, DEFAULT_REPORT_FILE);
         String maxResourceFileSizeStr = resolve(system, user, MAX_RESOURCE_FILE_SIZE, null);
         long maxResourceFileSize = DEFAULT_MAX_RESOURCE_FILE_SIZE;
@@ -426,6 +440,7 @@ public final class ScalpelConfiguration {
                 failSafe,
                 mode,
                 explain,
+                verifyFullBuild,
                 reportFile,
                 maxResourceFileSize,
                 warnings);
@@ -670,6 +685,40 @@ public final class ScalpelConfiguration {
     /** Returns whether per-module decision evidence is emitted (explain mode). Default: {@code false}. */
     public boolean isExplain() {
         return explain;
+    }
+
+    /**
+     * Returns whether the full build runs under observation and fails when a module Scalpel
+     * would have skipped fails (verify mode, #101). Default: {@code false}.
+     */
+    public boolean isVerifyFullBuild() {
+        return verifyFullBuild;
+    }
+
+    /**
+     * Returns a stable fingerprint of the configuration properties that shape the trim
+     * decision, for {@link ScalpelReport#computeDecisionId}. Output-shaping properties
+     * (reportFile, impactedLog, mode, failSafe, explain, ...) are deliberately excluded:
+     * the decision identity must not change because someone moved the report.
+     */
+    public String decisionFingerprint() {
+        StringBuilder sb = new StringBuilder();
+        appendFingerprintPart(sb, "baseBranch", baseBranch);
+        appendFingerprintPart(sb, "head", head);
+        appendFingerprintPart(sb, "alsoMake", String.valueOf(alsoMake));
+        appendFingerprintPart(sb, "alsoMakeDependents", String.valueOf(alsoMakeDependents));
+        appendFingerprintPart(sb, "fullBuildTriggers", String.join(",", fullBuildTriggers));
+        appendFingerprintPart(sb, "excludePaths", String.join(",", excludePaths));
+        appendFingerprintPart(sb, "includePaths", String.join(",", includePaths));
+        appendFingerprintPart(sb, "disableTriggers", String.join(",", disableTriggers));
+        appendFingerprintPart(sb, "forceBuildModules", String.join(",", forceBuildModules));
+        appendFingerprintPart(sb, "uncommitted", String.valueOf(uncommitted));
+        appendFingerprintPart(sb, "untracked", String.valueOf(untracked));
+        return sb.toString();
+    }
+
+    private static void appendFingerprintPart(StringBuilder sb, String name, String value) {
+        sb.append(name).append('=').append(value == null ? "" : value).append(';');
     }
 
     /** Returns the operating mode: {@code trim}, {@code skip-tests}, {@code report}, or {@code shadow}. Default: {@code trim}. */
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
@@ -707,14 +707,21 @@ public final class ScalpelConfiguration {
         appendFingerprintPart(sb, "head", head);
         appendFingerprintPart(sb, "alsoMake", String.valueOf(alsoMake));
         appendFingerprintPart(sb, "alsoMakeDependents", String.valueOf(alsoMakeDependents));
-        appendFingerprintPart(sb, "fullBuildTriggers", String.join(",", fullBuildTriggers));
-        appendFingerprintPart(sb, "excludePaths", String.join(",", excludePaths));
-        appendFingerprintPart(sb, "includePaths", String.join(",", includePaths));
-        appendFingerprintPart(sb, "disableTriggers", String.join(",", disableTriggers));
-        appendFingerprintPart(sb, "forceBuildModules", String.join(",", forceBuildModules));
+        appendFingerprintPart(sb, "fullBuildTriggers", sortedJoin(fullBuildTriggers));
+        appendFingerprintPart(sb, "excludePaths", sortedJoin(excludePaths));
+        appendFingerprintPart(sb, "includePaths", sortedJoin(includePaths));
+        appendFingerprintPart(sb, "disableTriggers", sortedJoin(disableTriggers));
+        appendFingerprintPart(sb, "forceBuildModules", sortedJoin(forceBuildModules));
+        appendFingerprintPart(sb, "maxResourceFileSize", String.valueOf(maxResourceFileSize));
         appendFingerprintPart(sb, "uncommitted", String.valueOf(uncommitted));
         appendFingerprintPart(sb, "untracked", String.valueOf(untracked));
         return sb.toString();
+    }
+
+    private static String sortedJoin(List<String> values) {
+        List<String> sorted = new ArrayList<>(values);
+        sorted.sort(java.util.Comparator.naturalOrder());
+        return String.join(",", sorted);
     }
 
     private static void appendFingerprintPart(StringBuilder sb, String name, String value) {
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
@@ -725,7 +725,19 @@ public final class ScalpelConfiguration {
     }
 
     private static void appendFingerprintPart(StringBuilder sb, String name, String value) {
-        sb.append(name).append('=').append(value == null ? "" : value).append(';');
+        // Values are hex-encoded: list-valued properties legally contain the structural
+        // delimiters (';' in paths, '=' in patterns), and a raw concatenation let two
+        // distinct resolved configurations collide into one fingerprint (#177 review).
+        sb.append(name).append('=').append(hex(value == null ? "" : value)).append(';');
+    }
+
+    private static String hex(String value) {
+        byte[] bytes = value.getBytes(java.nio.charset.StandardCharsets.UTF_8);
+        StringBuilder sb = new StringBuilder(bytes.length * 2);
+        for (byte b : bytes) {
+            sb.append("%02x".formatted(b));
+        }
+        return sb.toString();
     }
 
     /** Returns the operating mode: {@code trim}, {@code skip-tests}, {@code report}, or {@code shadow}. Default: {@code trim}. */
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
@@ -336,7 +336,10 @@ public final class ScalpelReport {
                 .append('|')
                 .append(configFingerprint)
                 .append('|');
-        buildSetPaths.stream().sorted().forEach(path -> canonical.append(path).append('\n'));
+        buildSetPaths.stream()
+                .sorted()
+                .forEach(path ->
+                        canonical.append(path.length()).append(':').append(path).append('\n'));
         try {
             java.security.MessageDigest digest = java.security.MessageDigest.getInstance("SHA-256");
             byte[] hash = digest.digest(canonical.toString().getBytes(StandardCharsets.UTF_8));
--- a/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
+++ b/core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
@@ -76,6 +76,7 @@ public final class ScalpelReport {
             Set.of(CATEGORY_DIRECT, CATEGORY_UPSTREAM, CATEGORY_DOWNSTREAM, CATEGORY_TRANSITIVE);
 
     private final String baseBranch;
+    private final String decisionId;
     private final String status;
     private final String reason;
     private final boolean fullBuildTriggered;
@@ -93,6 +94,7 @@ public final class ScalpelReport {
 
     private ScalpelReport(
             String baseBranch,
+            String decisionId,
             String status,
             String reason,
             boolean fullBuildTriggered,
@@ -108,6 +110,7 @@ public final class ScalpelReport {
             Timings timings,
             long totalMillis) {
         this.baseBranch = baseBranch;
+        this.decisionId = decisionId;
         this.status = status;
         this.reason = reason;
         this.fullBuildTriggered = fullBuildTriggered;
@@ -315,6 +318,39 @@ public final class ScalpelReport {
         }
     }
 
+    /**
+     * Computes the stable decision identity (#101): the SHA-256 hex digest of the merge-base
+     * commit, the head commit, the resolved decision-shaping configuration fingerprint and
+     * the module paths of the resulting build set, sorted so reactor order does not change
+     * the identity. Null git ids are tolerated (unmeasured runs hash the placeholder
+     * {@code "-"}). The id is emitted in every mode's report so a decision is quotable in a
+     * bug report and correlatable with a post-merge failure.
+     */
+    public static String computeDecisionId(
+            String mergeBaseId, String headId, String configFingerprint, Collection<String> buildSetPaths) {
+        StringBuilder canonical = new StringBuilder("v1|");
+        canonical
+                .append(mergeBaseId == null ? "-" : mergeBaseId)
+                .append('|')
+                .append(headId == null ? "-" : headId)
+                .append('|')
+                .append(configFingerprint)
+                .append('|');
+        buildSetPaths.stream().sorted().forEach(path -> canonical.append(path).append('\n'));
+        try {
+            java.security.MessageDigest digest = java.security.MessageDigest.getInstance("SHA-256");
+            byte[] hash = digest.digest(canonical.toString().getBytes(StandardCharsets.UTF_8));
+            StringBuilder hex = new StringBuilder(hash.length * 2);
+            for (byte b : hash) {
+                hex.append("%02x".formatted(b));
+            }
+            return hex.toString();
+        } catch (java.security.NoSuchAlgorithmException e) {
+            // SHA-256 is mandated for every Java platform; unreachable.
+            throw new IllegalStateException("SHA-256 unavailable", e);
+        }
+    }
+
     public String toJson() {
         StringBuilder sb = new StringBuilder();
         sb.append("{\n");
@@ -323,6 +359,9 @@ public final class ScalpelReport {
                 .append(jsonString(Version.version()))
                 .append(",\n");
         sb.append("  \"baseBranch\": ").append(jsonString(baseBranch)).append(",\n");
+        if (decisionId != null) {
+            sb.append("  \"decisionId\": ").append(jsonString(decisionId)).append(",\n");
+        }
         if (status != null) {
             sb.append("  \"status\": ").append(jsonString(status)).append(",\n");
         }
@@ -537,6 +576,7 @@ public final class ScalpelReport {
 
     public static class Builder {
         private String baseBranch;
+        private String decisionId;
         private String status;
         private String reason;
         private boolean fullBuildTriggered;
@@ -554,6 +594,11 @@ public final class ScalpelReport {
 
         public Builder baseBranch(String baseBranch) {
             this.baseBranch = baseBranch;
+            return this;
+        }
+
+        public Builder decisionId(String decisionId) {
+            this.decisionId = decisionId;
             return this;
         }
 
@@ -638,6 +683,7 @@ public final class ScalpelReport {
             }
             return new ScalpelReport(
                     baseBranch,
+                    decisionId,
                     status,
                     reason,
                     fullBuildTriggered,
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
@@ -26,6 +26,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
@@ -111,7 +112,7 @@ class ScalpelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
             List<String> selectedProjects = session.getRequest().getSelectedProjects();
             if (selectedProjects != null && !selectedProjects.isEmpty()) {
                 logger.info("Scalpel {} disabled due to -pl project selection", version);
-                if (config.isModeShadow() || config.isVerifyFullBuild()) {
+                if (monitoredRun(config)) {
                     // This exit predates any measurement; a previous run's shadow document
                     // must not survive it (#89 semantics, shadow twin).
                     writeShadowStatus(
@@ -145,7 +146,7 @@ class ScalpelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
             // Detect changes
             ChangeDetectionResult result = scalpelCore.detectChanges(reactorRoot, config, allPomPaths, timings);
             if (result == null) {
-                if (config.isPassiveMode() || config.isVerifyFullBuild()) {
+                if (passiveRun(config)) {
                     String skipReason = scalpelCore.getLastDetectionSkipReason();
                     if (skipReason != null) {
                         writeStatusReport(config, reactorRoot, "skipped", skipReason);
@@ -162,7 +163,7 @@ class ScalpelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
                 if (config.isBuildAllIfNoChanges()) {
                     logger.info("Scalpel: No changes detected, building all modules (buildAllIfNoChanges=true)");
                 }
-                if (config.isPassiveMode() || config.isVerifyFullBuild()) {
+                if (passiveRun(config)) {
                     writeStatusReport(config, reactorRoot, "skipped", "no changes detected");
                 }
                 return;
@@ -172,7 +173,7 @@ class ScalpelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
 
             // Check disable triggers (bail out entirely if any changed file matches)
             if (matchesDisableTrigger(changedFiles, config)) {
-                if (config.isPassiveMode() || config.isVerifyFullBuild()) {
+                if (passiveRun(config)) {
                     writeStatusReport(config, reactorRoot, "skipped", "disabled by disableTriggers match");
                 }
                 return;
@@ -182,7 +183,7 @@ class ScalpelLifecycleParticipant extends AbstractMavenLifecycleParticipant {
             changedFiles = filterExcludedPaths(changedFiles, config);
             if (changedFiles.isEmpty()) {
                 logger.info("Scalpel: All changed files excluded by path filters, building all modules");
-                if (config.isPassiveMode() || config.isVerifyFullBuild()) {
+                if (passiveRun(config)) {
                     writeStatusReport(config, reactorRoot, "skipped", "all changed files excluded by path filters");
                 }
                 return;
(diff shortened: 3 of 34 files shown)


core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
```diff
@@ -672,6 +687,47 @@ public boolean isExplain() {
...
+    }
+
+    private static String sortedJoin(List<String> values) {
+        List<String> sorted = new ArrayList<>(values);
+        sorted.sort(java.util.Comparator.naturalOrder());
+        return String.join(",", sorted);
+    }
+
+    private static void appendFingerprintPart(StringBuilder sb, String name, String value) {
+        sb.append(name).append('=').append(value == null ? "" : value).append(';');
```
gnodet: 🔴 **Bug: fingerprint encoding is ambiguous — produces real collisions.**

Semicolons (`;`) and equals signs (`=`) are structural delimiters in this encoding, but they are also legal characters in list-valued config properties (paths containing `;`, Ant-style patterns like `**/*.xml`). Two distinct configurations therefore produce the same fingerprint string and the same `decisionId`, silently correlating different decisions.

**Reproduced collision:**
```
// fullBuildTriggers=["x;excludePaths=y"], excludePaths=["z"]
// fullBuildTriggers=["x"],               excludePaths=["y;excludePaths=z"]
// Both encode to: fullBuildTriggers=x;excludePaths=y;excludePaths=z;
```

CodeRabbit flagged the same issue (line 728).

**Fix:** use a length-prefixed encoding to make field boundaries unambiguous:

```suggestion
    private static void appendFingerprintPart(StringBuilder sb, String name, String value) {
        String v = value == null ? "" : value;
        sb.append(name).append('=').append(v.length()).append(':').append(v).append(';');
    }
```

Also add a regression test: two configs with different `fullBuildTriggers`/`excludePaths` that currently collide must produce different fingerprints.

it/extension3-its/src/it/verify-success/invoker.properties
```diff
@@ -0,0 +1 @@
+invoker.goals = validate -Dscalpel.baseBranch=base -Dscalpel.verifyFullBuild=true
```
gnodet: 🟠 **Testing gap: `validate` doesn't run tests, making the happy-path assertion trivially true.**

`verifyFullBuild` is documented as "run the full build … all tests run". The `validate` phase compiles nothing and runs no tests. `ShadowBuildMonitor` is installed and the IT passes because no modules fail at `validate` — but that's not the same as proving the feature works correctly when tests actually run and pass.

The IT asserts `BUILD SUCCESS` and `"wouldHaveSkippedButFailed": []`, but both are trivially satisfied without any test execution. The verify-failure IT uses `verify` correctly. This fixture should too.

CodeRabbit flagged the same issue.

```suggestion
invoker.goals = verify -Dscalpel.baseBranch=base -Dscalpel.verifyFullBuild=true
```

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -653,8 +697,10 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
             if (config.isFailSafe()) {
                 logger.warn("Scalpel: {}, building all modules", e.getMessage());
                 logger.debug("ScalpelException details", e);
-                if (config.isModeShadow()) {
-                    // The monitor is the last thing the passive branch installs, so an
+                if (config.isModeShadow()
+                        || config
+                                .isVerifyFullBuild()) { // The monitor is the last thing the passive branch installs, so
```
gnodet: 🔵 **Nit: use `monitoredRun(config)` — that's why the helper was introduced.**

This is the only place in the diff that manually replicates `config.isModeShadow() || config.isVerifyFullBuild()` instead of delegating to the `monitoredRun()` helper extracted in this same PR. If a third monitored mode is ever added, this spot will silently miss it.

```suggestion
                if (monitoredRun(config)) {
```

(The trailing comment about the monitor should move to a preceding line or be dropped — it belongs to the surrounding context, not the condition.)

gnodet (dismissed):
Two issues need fixing before this can merge: a fingerprint collision that breaks the `decisionId` stability guarantee, and a verify-success IT that exercises the wrong lifecycle phase.

A third inconsistency (`monitoredRun` not used in one spot) is minor but should be fixed for consistency.

paoloantinori:
All three addressed in 5588897 (rebased onto the merged containment PR first):

**Fingerprint collision**: your reproducer became the RED test (`decisionFingerprint_valuesContainingDelimitersCannotCollide`, failing on the previous commit exactly as you showed), and the fix hex-encodes every fingerprint value so the structural delimiters can never appear inside one. The two colliding configurations now produce distinct fingerprints, and decisionIds with them.

**verify-success IT**: now runs the full `verify` lifecycle like the failure IT, so the happy-path assertions (BUILD SUCCESS, empty `wouldHaveSkippedButFailed`) are earned with tests actually executing.

**monitoredRun consistency**: that condition had been split across lines by a formatter and escaped the predicate extraction; it now delegates to `monitoredRun(config)` and the comment it had swallowed is back on its own lines.

Battery after the rebase: core 204, extension3 192 unit tests, all 38 ITs green.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
paoloantinori: Applied in 5588897 — see the PR comment for details.

it/extension3-its/src/it/verify-success/invoker.properties
paoloantinori: Applied in 5588897 — see the PR comment for details.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
paoloantinori: Applied in 5588897 — see the PR comment for details.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
```diff
@@ -315,6 +318,39 @@ public String getReason() {
...
+            String mergeBaseId, String headId, String configFingerprint, Collection<String> buildSetPaths) {
+        StringBuilder canonical = new StringBuilder("v1|");
+        canonical
+                .append(mergeBaseId == null ? "-" : mergeBaseId)
+                .append('|')
+                .append(headId == null ? "-" : headId)
+                .append('|')
+                .append(configFingerprint)
+                .append('|');
+        buildSetPaths.stream().sorted().forEach(path -> canonical.append(path).append('\n'));
```
gnodet: 🟠 **buildSetPaths newline ambiguity — same issue as the fingerprint collision, different location.**

Module paths are appended to the canonical string separated by `\n`, but on Linux a directory name can legally contain a newline. Two build sets that differ only in path-content-vs-separator interpretation produce the same bytes:

```
["module-a\nmodule-b"]  →  "module-a\nmodule-b\n"
["module-a", "module-b"]  →  "module-a\nmodule-b\n"
```

Identical canonical inputs → identical SHA-256 → same `decisionId` for distinct decisions. The same length-prefix encoding that fixed the fingerprint should be applied here. CodeRabbit flagged this in its last review and it remains open.

```suggestion
        buildSetPaths.stream().sorted().forEach(path -> {
            String p = path == null ? "" : path;
            canonical.append(p.length()).append(':').append(p).append('\n');
        });
```

Also add a regression test: `computeDecisionId(["a\nb"], ...) != computeDecisionId(["a", "b"], ...)`.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -191,7 +192,7 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
             // Check full build triggers
             String triggerFile = findFullBuildTrigger(changedFiles, config);
             if (triggerFile != null) {
-                if (config.isPassiveMode()) {
+                if (passiveRun(config)) {
                     writeFullBuildReport(config, reactorRoot, triggerFile, changedFiles);
```
gnodet: 🟡 **Full-build-trigger report missing `decisionId`.**

When a changed file matches a `fullBuildTriggers` pattern, `writeFullBuildReport` is called and the resulting report has no `decisionId`. Every other report path in this PR (all the `passiveRun` branches, the trim/shadow/verify paths) carries a `decisionId`. This one silently drops it, breaking the correlation contract the feature is specifically trying to establish: a scheduled `scalpel-verify` job would emit reports without an id when a trigger file fires.

CodeRabbit flagged this in its last review and it remains open.

`writeFullBuildReport` does not receive the `ChangeDetectionResult` it needs to call `decisionIdFor`. It needs to be threaded through, or a variant computed from the reactor projects (all modules, since the full build runs everything):

```java
// In the full-build-trigger branch (around line 194–196):
if (passiveRun(config)) {
    String fullBuildDecisionId = decisionIdFor(result, config, reactorRoot, allProjects);
    writeFullBuildReport(config, reactorRoot, triggerFile, changedFiles, fullBuildDecisionId);
}
```

And update `writeFullBuildReport` to accept and set the `decisionId` on the builder.

gnodet (commented):
Re-review of commit `5588897`.

All three issues from the previous `CHANGES_REQUESTED` review are addressed:
- ✅ Fingerprint collision — hex encoding + regression test (`decisionFingerprint_valuesContainingDelimitersCannotCollide`)
- ✅ verify-success fixture now runs `verify` instead of `validate`
- ✅ `monitoredRun(config)` consistency — the lone manual replication is gone

Two issues raised by CodeRabbit in its second review (on this same `5588897` commit) remain open. Both are in new code introduced by this PR.

paoloantinori:
The two open CodeRabbit findings are addressed in 5974033:

**Newline-in-path collision**: paths in the canonical hash were newline-separated, so a single path containing a literal newline hashed identically to the same content as two real paths. Each path is now length-prefixed in the canonical form; the forged-newline case became the RED test first (fails on 5588897, passes on 5974033).

**decisionId in full-build-trigger reports**: the full-build-trigger report now carries the id computed from the full reactor (the decision there being "build everything"), so a triggered run's report correlates with the shadow artifacts like every other mode.

Unit battery and the verify/shadow/full-build-trigger IT families are green.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
gnodet: Fixed in 5974033 — path encoding now uses length-prefix (`path.length():path`) to prevent newline ambiguity. Regression test added.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
gnodet: Fixed in 5974033 — full-build-trigger reports now compute the `decisionId` from the full reactor project list.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
gnodet: Fixed in 5974033 — path encoding now uses length-prefix (path.length():path) to prevent newline ambiguity. Regression test added.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
gnodet: Fixed in 5974033 — full-build-trigger reports now compute the decisionId from the full reactor project list.

gnodet (commented):
Re-review of commit `fb64731` (merge of `main` into `feat/verify-mode`).

The two issues from the previous `COMMENT` review (`5974033`) remain open. A third area — the report-mode `decisionId` inconsistency — was raised by CodeRabbit in its second review and is still present in the merged code.

---

### Previously raised findings

**From the `5974033` review (my COMMENT on that commit):**

No findings were raised by me on `5974033`; that review confirmed all three issues from the `REQUEST_CHANGES` review were fixed. The open items below come from CodeRabbit's second review (on `5588897`) which were **not** addressed by `5974033` and remain unaddressed in the current `fb64731` HEAD.

---

### Open findings

#### 1. Report-mode `decisionId` inconsistency — `allAffected` vs `directlyAffected`

`ScalpelLifecycleParticipant.java` lines 555–561: the plain-report branch computes `decisionProjects` from `trimResult.getBuildSet()`, where `trimResult` is computed from `directlyAffected` alone (line 482–490).

Shadow/verify mode at lines 513–546 computes the decision from `allAffected` (which includes transitively affected modules pulled in by POM/dependency analysis). For the same changeset, running in report mode produces a different `decisionId` than running in shadow/verify mode — which is exactly the cross-mode correlation problem the feature is supposed to solve.

CodeRabbit flagged this in its second review (inline comment on line 561). The fix: for the report-mode `decisionProjects`, use `allAffected` + trim-mode post-filtering (the same set shadow/verify computes), not `trimResult.getBuildSet()` from `directlyAffected`.

#### 2. Verify mode silent on empty decisions — no `ShadowBuildMonitor` installed when no modules are affected

`ScalpelLifecycleParticipant.java` lines 330–352: when `directlyAffected.isEmpty() && oldEffectiveModels.isEmpty()`, the code writes a shadow status document and returns. **No `ShadowBuildMonitor` is installed.** In verify mode (`scalpel.verifyFullBuild=true`), this means the full build runs unmonitored: every module is in fact "would-have-been-skipped" (nothing was affected), but if any module fails, the false negative goes undetected and the build exits success from Scalpel's perspective.

This was CodeRabbit's "Install the verify monitor for empty build decisions" finding (outside-diff comment, flagged as Major). The same gap exists at lines 397–420 (second `directlyAffected.isEmpty() && transitivelyAffected.isEmpty()` path) and 446–468 (the `allAffected.isEmpty()` after `includePaths` filtering).

The same `ShadowDecision.verifying(empty, allProjects, …)` pattern that is installed at line 607 for non-empty decisions should be installed at each of these early-return paths.

---

### False positives from CodeRabbit's second review

- **Surefire not pinned in `verify-failure/module-c/pom.xml`**: The IT runner downloads Maven 3.9.16 (`maven3Version`) which bundles `maven-surefire-plugin 3.5.5`. That version auto-discovers JUnit 5 via the JUnit Platform provider when `junit-jupiter` is present on the classpath. No explicit pin is needed — the JUnit 5 discovery path is unconditional in surefire 3.x. **Non-issue.**

- **Symlinks in `resolveContained`**: The flagged code is pre-existing in `main`, not introduced by this PR. **Out of scope.**

---

### Main-branch merge (`fb64731`)

The merge of `main` into `feat/verify-mode` brings in fixes #112, #113, #114, #115, #148, #96. None of these interact with the verify-mode logic in ways that introduce new issues. The `normalizedRoot` hoisting (#113) is clean. The BFS worklist for test-jar softening (#115) is correct.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
```diff
@@ -546,29 +526,91 @@ public void afterProjectsRead(MavenSession session) throws MavenExecutionExcepti
...
+                                config.getMode());
+                    }
+                } else {
+                    // Plain report run: the decision identity covers the set this report
+                    // describes (the enrichment trim result, or the affected set when the
+                    // reactor had nothing to enrich).
+                    Collection<MavenProject> decisionProjects = trimResult != null
+                            ? trimResult.getBuildSet()
+                            : concat(directlyAffected, transitivelyAffected.keySet());
+                    decisionId = decisionIdFor(result, config, reactorRoot, decisionProjects);
```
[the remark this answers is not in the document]
gnodet: Fixed in e5c22b5 — report-mode now computes the decision from `allAffected` + trim + includePaths filter, same logic as shadow/verify mode.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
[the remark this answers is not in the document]
gnodet: Partially addressed in e5c22b5 — full-build-trigger reports already got `decisionId` in 5974033. The empty-decision verify monitor is now installed at all three early-return paths.

gnodet: Fixed in e5c22b5 — the full-build-trigger report already got decisionId in 5974033; this commit fixes the report-mode branch to use allAffected for decisionId consistency.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
gnodet: Fixed in e5c22b5 — `installEmptyDecisionMonitor()` is now called at all three early-return paths. The helper is a no-op for pure shadow mode (preserving existing shadow-unaffected behavior), only installing the verifying monitor when `isVerifyFullBuild()` is true.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelConfiguration.java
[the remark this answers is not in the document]
gnodet: Fixed in 5588897 — hex encoding + regression test (`decisionFingerprint_valuesContainingDelimitersCannotCollide`).

it/extension3-its/src/it/verify-failure/module-c/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+    <!-- Independent of the change in module-b and of every other module, so trim would
+         skip it; its failing test is exactly what shadow mode must surface as a
+         would-have-skipped false negative. -->
+    <dependencies>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>5.13.4</version>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
gnodet: Not applicable — the IT runner uses Maven 3.9.16 (`maven3Version`) which bundles `maven-surefire-plugin 3.5.5`. Pinning it in the fixture POM would be redundant.

it/extension3-its/src/it/verify-success/invoker.properties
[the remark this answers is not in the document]
gnodet: Fixed in 5588897 — `invoker.goals` now uses `verify` instead of `validate`.

core/src/main/java/eu/maveniverse/maven/scalpel/core/ScalpelReport.java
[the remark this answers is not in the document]
gnodet: Fixed in 5974033 — path encoding now uses length-prefix (`path.length():path`) to prevent newline ambiguity. Regression test added.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/ScalpelLifecycleParticipant.java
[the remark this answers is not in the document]
gnodet: Addressed across two commits: `decisionId` for full-build-trigger reports in 5974033, verify monitor on empty decisions in e5c22b5.

gnodet (approved):
All review findings from the initial `CHANGES_REQUESTED` review, both CodeRabbit reviews, and the re-reviews have been addressed across commits `5588897`, `5974033`, and `e5c22b5`.

**CI status**: Windows unit tests (JDK 21/25) fail with 11 test failures — these are **pre-existing failures on `main`** (identical failures in [main run #34213204866](https://github.com/maveniverse/scalpel/actions/runs/34213204866)). All other CI jobs (Build, Verify matrix ×9, macOS/Ubuntu unit tests, CodeRabbit, Release Drafter) are green. Not a regression from this PR.

Ready to merge. 🚀

gnodet:
Pushed `25a15d8`: fix `ChangeFilter` glob matching on Windows.

The `ChangeFilter.accepts()` method used `Path.of(changePath)` with the default filesystem's `PathMatcher`. Since change paths contain Maven coordinates with `:` (e.g., `managedDependencies/com.foo:bar`), this throws `InvalidPathException` on Windows where `:` is illegal in file paths.

Replaced the `Path`/`PathMatcher` approach with regex-based glob matching that works cross-platform. This fixes the `changeFilter_excludeGlobStar` error and the `analyzeChanges_excludeManagedDependency` failure.

The remaining 9 `ScalpelLifecycleParticipantTest` Windows failures are **pre-existing on `main`** (same failures in [run 34213204866](https://github.com/maveniverse/scalpel/actions/runs/34213204866)) — they are not regressions from this PR.

gnodet (commented):
Re-review of commit `25a15d8`.

Context: all findings from the previous `COMMENT` reviews on `e5c22b5` and `fb64731` have been addressed:
- ✅ Empty-decision verify monitor installed at all three early-return paths (`installEmptyDecisionMonitor`) — `e5c22b5`
- ✅ Report-mode `decisionId` consistency (uses `allAffected` + trim + includePaths) — `e5c22b5`
- ✅ Path traversal via symlinks on `resolveContained` — CodeRabbit finding; pre-existing in `main`, not introduced by this PR, outside scope

The new commit `25a15d8` replaces the `PathMatcher`-based `ChangeFilter` with a pure-regex implementation to avoid `InvalidPathException` on Windows where change paths like `managedDependencies/com.foo:bar` contain `:`. The fix is correct for the problem it solves. Two observations:

**`**/suffix` edge case in `globToRegex`:** `**/suffix` translates to `.*suffix` (the `/` separator is consumed from the pattern but not required in the match). This causes `.*suffix` to match the bare string `suffix` (without a preceding slash segment), which is not standard glob semantics. In practice all Scalpel change paths have a category prefix (`properties/`, `dependencies/`, `managedDependencies/`, etc.), so a real user pattern like `**/something` would never encounter a bare `something` path — the production risk is nil. Noting it because a future `globToRegex` unit test should document this and either fix it (e.g., `(.*/)?suffix`) or explicitly accept it.

**No direct unit test for `globToRegex`:** The behavioral `ChangeFilter` tests cover the common patterns well (`managedDependencies/**`, `properties/build.*`, include-override-exclude). The translation function itself is `static` and testable directly; a test named `globToRegex_translatesStarStarSlash` would make future changes to the regex logic safer. Minor — does not block merge.

**Merge conflict:** CodeRabbit's pre-merge checks flag an unresolved merge conflict in `feat/verify-mode`. This needs resolving before merge (rebase or merge `origin/main`).

Overall: the main body of the PR (`verifyFullBuild` mode, stable `decisionId`, new ITs) is solid and all previously raised findings are addressed. The `25a15d8` Windows fix is correct. The branch needs a rebase/conflict resolution, then it is ready.

gnodet (approved):
Re-review of commit `f6322f2` (merge of `origin/main` into `feat/verify-mode`).

All three observations from the previous `COMMENT` review on `25a15d8` are addressed:

- ✅ **`**/suffix` edge case** — the new `globToPattern` emits `(.*/)?\.suffix` instead of the old `.*suffix`. `**/pom.xml` now correctly rejects `parentpom.xml` (no path separator before the suffix). The `globToPattern_doubleStarSlashRequiresSeparator` test documents and validates this.

- ✅ **Direct `globToPattern` unit tests** — 5 new tests covering: `**/` prefix, `a/**/b` middle, trailing `**`, mid-segment `**` fallback to single-segment, and ReDoS safety (`**a**a**z` pattern completes in < 1000 ms on a 400-char input). The ReDoS-safety guarantee now comes from `[^/]*` rather than unbounded `.*` in mid-segment position.

- ✅ **Merge conflict resolved** — clean merge commit.

**What the merge brings in:**
- `ReactorFirstModelResolver.addRepository` now intentionally no-ops instead of delegating to a `RemoteRepositoryManager`-dependent delegate (correct: standalone model resolver doesn't have remote repo infrastructure).
- POSIX-permission tests guarded with `assumeTrue(… posix …)` so they skip on Windows NTFS.
- `\r\n` normalization in the golden-report equality assertion.
- Improved IT assertions for `skip-tests-downstream-exclusion-report`: uses `JsonSlurper` instead of brace-counting, adds missing `testsSkipped == true` assertion for module-b, and uses `containsKey` for module-c negative assertions.
- Windows temp-directory fallback (`TEMP` / `TMP` / `TMPDIR`) in `detectChanges_notAGitRepo_returnsNull`.

All CI jobs green including Windows (unit 21/25, verify matrix ×15 including `windows-latest`). The verify-mode logic is untouched by the merge. Ready to merge. 🚀

gnodet (commented):
Re-review of commit `7b702bf` (merge of `origin/main` into `feat/verify-mode`).

The merge conflict in `ScalpelLifecycleParticipant.java` caused by the #174 refactoring (extract-collaborators) has been resolved.

**Adaptation summary:**
- `passiveRun`, `monitoredRun`, `installEmptyDecisionMonitor`, `moduleKeyOf`, `decisionIdFor` helpers ported onto the refactored participant
- All direct method calls replaced with collaborator APIs (`reportAssembler.*`, `changedFileClassifier.*`, etc.)
- `ChangesetContext` gains a `result` field so `handleNoAffectedModules` can compute the `decisionId` for empty-decision reports
- `ReportAssembler.writeFullBuildReport` updated: new `decisionId` parameter + verify-mode shadow write
- `ReportAssembler.writeReport` and `writeStatusReport` updated for verify-mode consistency
- `installEmptyDecisionMonitor` called at all three `handleNoAffectedModules` early-return paths

All three observations from the previous `COMMENT` review on `25a15d8` remain addressed; this merge commit brings in `origin/main`'s refactoring without regressions.

gnodet (commented):
Re-review triggered by NEW_COMMENTS signal — no new comments found since the last review.

**Status update:**

The branch now conflicts with `main` again. Two commits merged since the last `origin/main` integration (`f6322f2`):

1. **`ece74a1`** — Fix #125: extract single-responsibility collaborators from `ScalpelLifecycleParticipant` (#174). This is the same refactoring that the previous merge commit (`f6322f2`) adapted to, but the actual PR is now on `main` with its final form — causing 16 conflict regions in `ScalpelLifecycleParticipant.java`.

2. **`59e2347`** — Fix #116: replace per-project DFS with precomputed BFS traversals. Touches `PomChangeAnalyzer`, `ReactorTrimmer`, and `ReactorTrimmerTest` — no direct conflict with this PR's files, but the merge machinery trips over the combined diff.

**CI**: all 18 checks green on the current HEAD (`f6322f2`). No regressions.

**Code review**: all previously raised findings remain addressed. No new issues.

**Next step**: the branch needs a rebase or merge of `origin/main` to resolve the #174 conflict. The conflict is entirely in `ScalpelLifecycleParticipant.java` where the verify-mode additions need to be ported onto the post-#174 collaborator structure (which the PR already did once in `f6322f2` — this is essentially a replay).

gnodet (commented):
Re-review triggered by NEW_COMMENTS signal — no new comments found since the last review (`2026-09-09T08:28:49Z`).

**Status unchanged from the previous review:**

- **HEAD**: `f6322f2` (merge of `origin/main` into `feat/verify-mode`)
- **CI**: all 18 checks green (including Windows unit tests JDK 21/25 and the full 9-matrix verify suite)
- **Code**: all previously raised findings remain addressed
- **Merge conflict**: branch still conflicts with `main`. Two commits merged since the last `origin/main` integration: #174 (extract collaborators) and #116 (BFS traversals). The conflict is in `ScalpelLifecycleParticipant.java` — the verify-mode additions need to be ported onto the post-#174 collaborator structure (same adaptation already done once in `f6322f2`).

**Approval**: the previous APPROVE on `f6322f2` stands. The PR needs a rebase/merge of `origin/main` to resolve the conflict before it can be merged.

gnodet (commented):
Re-review of commit `9f1d55b` (empty re-trigger commit after merge of `origin/main` into `feat/verify-mode`).

The previous merge commit `7b702bf` did not trigger CI (GitHub Actions glitch — the run was never queued despite the push event). Pushed an empty commit to re-trigger.

**Mergeability**: GitHub's `mergeable_state` was stuck at `dirty` for `7b702bf` even though the branch is a clean superset of `main` (fast-forward merge verified locally). Now cleared — `mergeable: true`.

**CI**: queued and running on `9f1d55b`. The diff from `main` is identical to the reviewed `7b702bf` merge (only an empty commit added). Awaiting green.

**Review status**: all findings addressed across the review cycle. Two APPROVALs already posted (`e5c22b5` at 11:54 and `f6322f2` at 18:58). No new code changes since the last APPROVED review — the only additions are merge commits bringing in `main`'s #174 and #116.

Merge once CI is green. 🚀

gnodet (commented):
Re-review triggered by NEW_COMMENTS signal — no new comments or code changes since the last review.

**Status update:**

- **HEAD**: `9f1d55b` (empty CI re-trigger commit on top of the `7b702bf` merge)
- **Merge conflict**: ✅ resolved — branch is now `MERGEABLE`
- **CI**: running on `9f1d55b` (all checks in progress)
- **Code**: unchanged since `7b702bf` — all previously raised findings remain addressed

The previous APPROVE on `f6322f2` stands. Waiting for CI to go green on the new HEAD, then ready to merge.

gnodet (commented):
Re-review triggered by NEW_COMMENTS signal — checking commits since last review (`f6322f2`).

**New commits on `feat/verify-mode`:**

1. **`7b702bf`** — Merge of `origin/main` into `feat/verify-mode`. Resolves the 16-region conflict in `ScalpelLifecycleParticipant.java` caused by #174 (extract collaborators) and #116 (BFS traversals). The verify-mode additions (`passiveRun`, `monitoredRun`, `installEmptyDecisionMonitor`, `moduleKeyOf`, `decisionIdFor`) are ported onto the refactored collaborator structure. `ChangesetContext` gains a `result` field for empty-decision `decisionId` computation. `ReportAssembler` updated for verify-mode shadow writes and `decisionId` parameter.

2. **`9f1d55b`** — Empty commit (`ci: re-trigger CI after merge`). No code changes.

**Merge conflict**: ✅ resolved. Branch merge-base is `59e2347` (tip of `main`). GitHub reports `MERGEABLE`.

**CI**: running now on `9f1d55b` (all jobs in_progress — Build, Unit tests ×6, Release Drafter). Previous CI on `f6322f2` was all-green (18/18 checks pass including Windows). Since `9f1d55b` is an empty commit on top of `7b702bf` which is a clean merge, expecting the same result.

**Code review**: all previously raised findings remain addressed:
- ✅ Fingerprint collision — hex encoding + regression test
- ✅ verify-success IT runs `verify` phase
- ✅ `monitoredRun(config)` consistency
- ✅ Newline-in-path collision — length-prefixed
- ✅ `decisionId` in full-build-trigger reports
- ✅ Empty-decision verify monitor at all three early-return paths
- ✅ Report-mode `decisionId` consistency (`allAffected` + trim + includePaths)
- ✅ `globToPattern` `**/suffix` edge case fixed + 5 direct unit tests + ReDoS safety
- ✅ Windows `ChangeFilter` glob matching (pure-regex, no `Path.of`)

**No new comments from the contributor since the last review.** The only activity is the conflict resolution merge and CI re-trigger.

**Approval**: the previous APPROVEs on `f6322f2` stand. Waiting for CI on `9f1d55b` to go green, then ready to merge.

gnodet:
Re-review of commit `9f1d55b` (merge of `origin/main` + CI re-trigger).

The merge conflict caused by #174 (extract-collaborators) and #116 (BFS traversals) landing on `main` has been resolved in `7b702bf`. The verify-mode additions have been ported onto the post-#174 collaborator structure. No code changes beyond the merge resolution.

**CI**: running now on `9f1d55b` — 7 checks in progress, release-drafter ✅, sonarcloud skipped. Will monitor for results.

**Review status**: all previously raised findings remain addressed. No new comments from the PR author or other reviewers since the last review.

gnodet (approved):
Re-review of commits `7b702bf` (merge of `origin/main` into `feat/verify-mode`) and `9f1d55b` (CI re-trigger).

The merge conflict caused by #174 (extract-collaborators) and #116 (BFS traversals) landing on `main` has been correctly resolved. The verify-mode logic has been properly ported onto the post-#174 collaborator structure:

- ✅ `passiveRun()` / `monitoredRun()` helpers present and used at all guard sites
- ✅ `installEmptyDecisionMonitor()` called at all three `handleNoAffectedModules` early-return paths
- ✅ `moduleKeyOf()` / `decisionIdFor()` helpers ported and used for consistent module naming
- ✅ `ChangesetContext` record gains `result` field for `decisionId` computation in `handleNoAffectedModules`
- ✅ `ReportAssembler.writeFullBuildReport` updated: `decisionId` parameter + verify-mode shadow status write
- ✅ `ReportAssembler.writeStatusReport`: handles verify mode (`isVerifyFullBuild`) alongside shadow
- ✅ `ReportAssembler.writeReport`: passes `ctx.decisionId` to `ScalpelReport` builder via `AnalysisContext`
- ✅ `AnalysisContext` / `AnalysisContext.empty()` updated to carry `decisionId`
- ✅ Report-mode `decisionId` consistency (uses `allAffected` + trim + `includePaths` filter) preserved in the ported code
- ✅ Trim-mode `decisionIdFor` call added before `reportAssembler.writeReport` in the trim branch

No new code issues introduced by the merge resolution. All previously raised findings remain addressed.

**CI**: unit tests (6/6) ✅, Build ✅, Verify matrix (9 jobs) still running. Previous approval on `f6322f2` stands.

gnodet:
✅ **CI is green** — all 18 check runs on commit `9f1d55b` have completed successfully (17 passed, 1 skipped: sonarcloud). This PR is ready to merge.

PR #178, opened 2026-09-08
test: parse the skip-tests-downstream-exclusion-report IT instead of brace-slicing (#96)

paoloantinori:
The follow-up accepted in https://github.com/maveniverse/scalpel/pull/170#issuecomment-5574803747: the last remaining hand-sliced JSON assertions from issue #96.

The `skip-tests-downstream-exclusion-report` verify script extracted module objects with `lastIndexOf('{')`/`indexOf('}')`, the exact fragility the issue names, since the first nested field inside a module object would silently corrupt those extractions. It now parses with `JsonSlurper` like the other report ITs and asserts per-module fields by artifactId; the module-b `testsSkippedReason=EXCLUDED_DOWNSTREAM` and the module-c absent-field checks run against the typed structure, which is strictly stronger than substring matches on hand-sliced blocks.

Unit-side coverage (escaping tests with raw-output verification, SLPTest substring helpers replaced) ships in #170; this closes the one IT the issue text calls out. Green locally.

## Summary by CodeRabbit

* **Tests**
  * Improved verification of downstream exclusion reports.
  * Added explicit checks for module change reasons, categories, and skipped-test details.
  * Confirmed that unaffected modules omit skipped-test fields entirely.

### DIFF
--- a/it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
+++ b/it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
@@ -30,20 +30,26 @@ def report = new groovy.json.JsonSlurper().parseText(reportFile.text)
 assert report.version == '2' : "Report should have version 2"
 assert report.fullBuildTriggered == false : "fullBuildTriggered should be false"
 
-def modules = report.affectedModules.collectEntries { [(it.artifactId): it] }
+def modules = report.affectedModules
+def moduleA = modules.find { it.artifactId == 'module-a' }
+def moduleB = modules.find { it.artifactId == 'module-b' }
+def moduleC = modules.find { it.artifactId == 'module-c' }
 
 // module-a should be directly affected with SOURCE_CHANGE
-assert modules['module-a'] : "module-a should appear in report"
-assert modules['module-a'].reasons.contains('SOURCE_CHANGE') : "module-a should have SOURCE_CHANGE reason"
-assert modules['module-a'].category == 'DIRECT' : "module-a should have DIRECT category"
+assert moduleA : "module-a should appear in report"
+assert moduleA.reasons.contains('SOURCE_CHANGE') : "module-a should have SOURCE_CHANGE reason"
+assert moduleA.category == 'DIRECT' : "module-a should have DIRECT category"
 
-// module-b should be downstream with testsSkippedReason=EXCLUDED_DOWNSTREAM
-assert modules['module-b'] : "module-b should appear in report"
-assert modules['module-b'].reasons.contains('DOWNSTREAM_DEPENDENT') : "module-b should have DOWNSTREAM_DEPENDENT reason"
-assert modules['module-b'].testsSkippedReason == 'EXCLUDED_DOWNSTREAM' : \
+// module-b should be downstream with testsSkipped + testsSkippedReason=EXCLUDED_DOWNSTREAM
+assert moduleB : "module-b should appear in report"
+assert moduleB.reasons.contains('DOWNSTREAM_DEPENDENT') : "module-b should have DOWNSTREAM_DEPENDENT reason"
+assert moduleB.testsSkipped == true : "module-b should have testsSkipped=true"
+assert moduleB.testsSkippedReason == 'EXCLUDED_DOWNSTREAM' : \
     "module-b should have testsSkippedReason=EXCLUDED_DOWNSTREAM"
 
-// module-c should be downstream but WITHOUT testsSkippedReason
-assert modules['module-c'] : "module-c should appear in report"
-assert !modules['module-c'].containsKey('testsSkippedReason') : \
+// module-c should be downstream but WITHOUT testsSkipped/testsSkippedReason
+assert moduleC : "module-c should appear in report"
+assert !moduleC.containsKey('testsSkipped') : \
+    "module-c should NOT have testsSkipped in report"
+assert !moduleC.containsKey('testsSkippedReason') : \
     "module-c should NOT have testsSkippedReason in report"
--- a/it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
+++ b/it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
@@ -23,30 +23,27 @@ assert log.contains('BUILD SUCCESS')
 File reportFile = new File(basedir, 'target/scalpel-report.json')
 assert reportFile.exists() : "Report file should have been created at target/scalpel-report.json"
 
-String json = reportFile.text
-assert json.contains('"version": "2"') : "Report should contain version field"
-assert json.contains('"fullBuildTriggered": false') : "fullBuildTriggered should be false"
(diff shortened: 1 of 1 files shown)


it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
```diff
@@ -23,30 +23,27 @@ assert log.contains('BUILD SUCCESS')
...
-assert json.contains('"version": "2"') : "Report should contain version field"
-assert json.contains('"fullBuildTriggered": false') : "fullBuildTriggered should be false"
+// Parsed, not hand-sliced: brace-counting breaks the moment a module object
+// gains a nested field, which is exactly how a schema addition should NOT
+// fail these assertions (see #96).
+def report = new groovy.json.JsonSlurper().parseText(reportFile.text)
+assert report.version == '2' : "Report should have version 2"
+assert report.fullBuildTriggered == false : "fullBuildTriggered should be false"
+
+def modules = report.affectedModules.collectEntries { [(it.artifactId): it] }
```
gnodet: 🔶 **Convention:** Every other IT in `extension3-its/src/it/` navigates the parsed structure with `modules.find { it.artifactId == 'X' }` (see `bom-import`, `profile-aware`, `report-transitive`, `test-only-changes`, `resource-filtering`). This PR is the only one using `collectEntries` to build a lookup map. The map works fine for this fixture (unique artifactIds), but it diverges from the established pattern without a reason and silently drops duplicate entries if they ever appear.

Suggest aligning with the codebase convention:

```suggestion
def modules = report.affectedModules
def moduleA = modules.find { it.artifactId == 'module-a' }
def moduleB = modules.find { it.artifactId == 'module-b' }
def moduleC = modules.find { it.artifactId == 'module-c' }
```

(Then replace `modules['module-x']` → `moduleX` in the assertions below.)

it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
```diff
@@ -23,30 +23,27 @@ assert log.contains('BUILD SUCCESS')
...
-// Parse module-b block and verify testsSkippedReason
-def moduleBStart = json.indexOf('"module-b"')
-assert moduleBStart >= 0
-def moduleBBlock = json.substring(json.lastIndexOf('{', moduleBStart), json.indexOf('}', moduleBStart) + 1)
-assert moduleBBlock.contains('"testsSkippedReason": "EXCLUDED_DOWNSTREAM"') : \
-    "module-b should have testsSkippedReason=EXCLUDED_DOWNSTREAM in report, got: $moduleBBlock"
+assert modules['module-b'] : "module-b should appear in report"
+assert modules['module-b'].reasons.contains('DOWNSTREAM_DEPENDENT') : "module-b should have DOWNSTREAM_DEPENDENT reason"
+assert modules['module-b'].testsSkippedReason == 'EXCLUDED_DOWNSTREAM' : \
+    "module-b should have testsSkippedReason=EXCLUDED_DOWNSTREAM"
```
gnodet: 🔶 **Coverage gap:** `ScalpelReport.appendModuleJson` emits `"testsSkipped": true` as a companion boolean *before* `testsSkippedReason` (see `ScalpelReport.java` `appendModuleJson`). The test verifies `testsSkippedReason` but never asserts the boolean flag, leaving half the contract untested. Add:

```suggestion
assert modules['module-b'].testsSkipped == true : "module-b should have testsSkipped=true"
assert modules['module-b'].testsSkippedReason == 'EXCLUDED_DOWNSTREAM' : \
    "module-b should have testsSkippedReason=EXCLUDED_DOWNSTREAM"
```

gnodet (commented):
The substring→parser migration is the right fix for #96, and the assertions themselves are correct. Two nits and one consistency issue worth addressing.

it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
```diff
@@ -23,30 +23,27 @@ assert log.contains('BUILD SUCCESS')
...
 // module-c should be downstream but WITHOUT testsSkippedReason
-assert json.contains('"module-c"') : "module-c should appear in report"
-def moduleCStart = json.indexOf('"module-c"')
-assert moduleCStart >= 0
-def moduleCBlock = json.substring(json.lastIndexOf('{', moduleCStart), json.indexOf('}', moduleCStart) + 1)
-assert !moduleCBlock.contains('testsSkippedReason') : \
-    "module-c should NOT have testsSkippedReason in report, got: $moduleCBlock"
+assert modules['module-c'] : "module-c should appear in report"
+assert !modules['module-c'].containsKey('testsSkippedReason') : \
+    "module-c should NOT have testsSkippedReason in report"
```
gnodet: 🔶 **Coverage gap:** Since `testsSkipped` and `testsSkippedReason` are always emitted together, the absent-field check for module-c should cover both:

```suggestion
assert !modules['module-c'].containsKey('testsSkipped') : \
    "module-c should NOT have testsSkipped in report"
assert !modules['module-c'].containsKey('testsSkippedReason') : \
    "module-c should NOT have testsSkippedReason in report"
```

it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
gnodet: Fixed in 86475ab10a3705cc51e7635a54dc0177ea240e22 — switched to `.find { it.artifactId == '...' }` pattern, consistent with the other ITs.

it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
gnodet: Fixed in 86475ab10a3705cc51e7635a54dc0177ea240e22 — added `testsSkipped == true` assertion for module-b.

it/extension3-its/src/it/skip-tests-downstream-exclusion-report/verify.groovy
gnodet: Fixed in 86475ab10a3705cc51e7635a54dc0177ea240e22 — added `!containsKey('testsSkipped')` assertion for module-c.

gnodet (approved):
All three findings from the previous review are fully addressed in 86475ab:

- **Convention** ✅ — switched from `collectEntries` map to `modules.find { it.artifactId == '...' }`, consistent with every other IT in the suite.
- **Coverage gap (module-b)** ✅ — `assert moduleB.testsSkipped == true` added; full boolean+reason contract is now tested.
- **Coverage gap (module-c)** ✅ — replaced the misleading `testsSkippedReason == null` check (which would pass even if the key was present with null) with `!containsKey('testsSkipped')` and `!containsKey('testsSkippedReason')` — semantically correct and strictly stronger.

CodeRabbit re-review: no actionable comments. Semgrep: 0 findings. Ready to merge.

PR #179, opened 2026-09-08
Fix #106: run unit tests on real Windows and macOS runners, correct the Java claim

### LINKED ISSUE #106: [ci] CI OS matrix never actually runs Windows/macOS (parent workflow runs-on is hardcoded); unit tests run on one JDK only
## Problem

In `maveniverse/parent`'s reusable ci workflow (the release-55 one Scalpel delegates to), the `test` job declares an `os` matrix axis but the job header is a hardcoded `runs-on: ubuntu-latest`; the matrix value is used only in the artifact name. The Windows and macOS legs are duplicate Ubuntu runs. This is a parent-repo bug, but Scalpel inherits it, and it is precisely why path-separator defects survive undetected in a tool whose core operation is comparing paths. (I can file it against the parent repo instead if preferred.)

Second, independent gap: the `maven-test` invocation uses `-rf :it`, so the ~7,500 lines of unit tests run only once, in the build job on ubuntu/JDK 21. Meanwhile the README claims "Java: 8+" and `maven.compiler.release=8`, but no Java 8 runtime proof exists anywhere. The `-rf :it` shape is forced by `testRelease=21` + text blocks, so the fix is a separate job, not a wider matrix. Baseline for context: 216 unit tests, ~3s surefire time; there is no speed reason not to run them widely.

## Proposal

- Real per-OS runners, either fixed in the parent workflow or via a Scalpel-local job.
- Unit tests executing on Windows and macOS, plus at least two JDKs; exercise the extension at runtime on Java 8 or correct the claim.

paoloantinori:
The reusable parent workflow's test job hardcodes `runs-on: ubuntu-latest` (the `os-matrix` input only feeds artifact names), so Scalpel's Windows and macOS legs have been duplicate Ubuntu runs and path-separator defects had no leg that could catch them, in a tool whose core operation is comparing paths. Until the parent is fixed, Scalpel adds its own unit-test matrix: **three operating systems x JDK 21 and 25**, six legs of ~3s unit tests, with the heavy verify-plus-ITs staying in the build job. All actions pinned to commit SHAs per the repo's own #102 policy.

The README claim `Java: 8+` was stale: the JDK 17 bump made `maven.compiler.release` 17, so the extension has required Java 17+ bytecode since then and no Java 8 runtime could load it. The claim now matches reality instead of promising an untested, impossible floor (the issue's alternative: exercise Java 8 at runtime or correct the claim).

**Verified before pushing** (gate agent, clean clone with an empty local repository): the exact matrix invocation builds the full reactor green from scratch on JDK 21 and 25 (200 core + 190 extension3 tests); the `it/*` modules are inert under `test` since invoker lives in the `run-its` profile. One blocking finding from that verification, applied: JDK 17 legs were dropped because the parent-62 enforcer requires build-time Java `[21,)` and every 17 leg would die at validate before running a test, which is why the existing CI's 17 entry only ever ran the `-rf :it` tail. Two hardenings from the same gate: the cache pin comment names v4.3.0, and `.gitattributes` pins `mvnw` to LF so a CRLF-default Windows checkout cannot break the wrapper under bash.

**Deliberate choices to note:** the matrix deliberately stops at unit tests (the IT matrix would multiply ~40 invoker builds by six legs); the parent-repo `runs-on` bug itself is left for the parent project, per the issue's framing.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 .mvn/nisse.properties export-subst
+mvnw text eol=lf
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,10 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest, macOS-latest ]
-        java: [ 17, 21, 25 ]
+        # The build itself requires JDK 21+ (parent-62 enforcer
+        # requireBuildtimeJavaVersion.range=[21,)); 17 is only the
+        # bytecode floor. Two build JDKs satisfy the issue's ask.
+        java: [ 21, 25 ]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Set up JDK ${{ matrix.java }}
@@ -40,7 +43,7 @@ jobs:
           java-version: ${{ matrix.java }}
           distribution: zulu
       - name: Cache Maven repository
-        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
         with:
           path: ~/.m2/repository
           key: ${{ runner.os }}-maven-${{ matrix.java }}-${{ hashFiles('**/pom.xml') }}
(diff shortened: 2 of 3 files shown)


gnodet:
Closing as duplicate of #172 which incorporates the best parts of this PR (SHA-pinned actions, .gitattributes mvnw LF, README Java 17+ correction) along with the actual Windows test fixes needed to make CI green. Thanks @paoloantinori!

PR #180, opened 2026-09-08
Fix Windows CI test failures in ChangeFilter and ReactorFirstModelResolver

gnodet:
## Problem

Windows CI fails with 11 test failures (2 in PomChangeAnalyzerTest, 9 in ScalpelLifecycleParticipantTest).

### Root Cause 1: ChangeFilter uses java.nio.file.Path for logical paths

`ChangeFilter.accepts()` calls `Path.of(changePath)` on strings like `managedDependencies/com.foo:bar`. On Windows, `:` is illegal in file paths → `InvalidPathException`.

**Fix:** Replace `PathMatcher`/`Path.of()` with a simple glob-to-regex converter. The glob syntax (`*`, `**`, `?`) is converted to regex patterns at construction time, then matched against the string directly. No filesystem involvement.

### Root Cause 2: ReactorFirstModelResolver corrupts delegate state

`ReactorFirstModelResolver.addRepository()` delegated to the wrapped `ProjectModelResolver`, which calls `RemoteRepositoryManager.aggregateRepositories()`. During standalone effective model building (for old/new POM comparison), the `RemoteRepositoryManager` may not be fully configured, causing the delegate's internal repository list to become null and subsequent `resolveModel` calls to fail. This made `buildEffectiveModels()` return empty maps, causing all parent-POM analysis to fall back to conservative "mark everything affected" behavior.

**Fix:** Made `addRepository()` a no-op in `ReactorFirstModelResolver`. The reactor-first resolver only resolves from reactor POMs; it never needs remote repositories, so absorbing repository declarations is both safe and correct.

## Testing

All 221 extension3 tests pass on Linux. The fixes are cross-platform by design — no OS-specific checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved change-path filtering consistency across operating systems, including support for nested and root-level glob patterns.
  * Repository declarations are now excluded from delegated resolution.

* **Tests**
  * Added coverage for wildcard matching across path roots, nested segments, and trailing patterns.

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
@@ -14,13 +14,10 @@ import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.nio.file.FileSystem;
-import java.nio.file.FileSystems;
 import java.nio.file.FileVisitOption;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.PathMatcher;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.nio.file.attribute.FileAttribute;
@@ -35,6 +32,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
+import java.util.regex.Pattern;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
@@ -336,15 +334,12 @@ class PomChangeAnalyzer {
      * <p>When both lists are empty, all changes are accepted (the filter is a no-op).
      */
     static class ChangeFilter {
-        private static final String GLOB_PREFIX = "glob:";
-
-        private final List<PathMatcher> excludeMatchers;
-        private final List<PathMatcher> includeMatchers;
+        private final List<Pattern> excludeMatchers;
+        private final List<Pattern> includeMatchers;
 
         ChangeFilter(List<String> excludePatterns, List<String> includePatterns) {
-            FileSystem fs = FileSystems.getDefault();
-            this.excludeMatchers = compileMatchers(excludePatterns, fs);
-            this.includeMatchers = compileMatchers(includePatterns, fs);
+            this.excludeMatchers = compileMatchers(excludePatterns);
+            this.includeMatchers = compileMatchers(includePatterns);
         }
 
         /** Returns {@code true} if the change path should be considered (not excluded). */
@@ -352,15 +347,14 @@ class PomChangeAnalyzer {
             if (excludeMatchers.isEmpty() && includeMatchers.isEmpty()) {
                 return true;
             }
-            Path path = Path.of(changePath);
             // Include overrides exclude
-            for (PathMatcher m : includeMatchers) {
-                if (m.matches(path)) {
+            for (Pattern m : includeMatchers) {
+                if (m.matcher(changePath).matches()) {
                     return true;
                 }
             }
-            for (PathMatcher m : excludeMatchers) {
-                if (m.matches(path)) {
+            for (Pattern m : excludeMatchers) {
+                if (m.matcher(changePath).matches()) {
                     return false;
                 }
             }
@@ -371,15 +365,51 @@ class PomChangeAnalyzer {
             return !excludeMatchers.isEmpty() || !includeMatchers.isEmpty();
         }
 
-        private static List<PathMatcher> compileMatchers(List<String> patterns, FileSystem fs) {
+        private static List<Pattern> compileMatchers(List<String> patterns) {
             if (patterns == null || patterns.isEmpty()) {
                 return List.of();
             }
-            List<PathMatcher> matchers = new ArrayList<>(patterns.size());
+            List<Pattern> matchers = new ArrayList<>(patterns.size());
             for (String pattern : patterns) {
-                matchers.add(fs.getPathMatcher(GLOB_PREFIX + pattern));
+                matchers.add(globToPattern(pattern));
             }
             return matchers;
+        }
+
+        /**
+         * Converts a simple glob pattern to a compiled {@link Pattern}.
+         * The glob syntax supports: {@code *} (any chars except {@code /}),
+         * {@code **} (any chars including {@code /}), {@code ?} (single char).
+         * Forward-slash {@code /} is the separator.  No filesystem is involved,
+         * so characters like {@code :} that are illegal in Windows file paths
+         * are handled correctly.
+         */
+        static Pattern globToPattern(String glob) {
+            StringBuilder regex = new StringBuilder();
+            int i = 0;
+            while (i < glob.length()) {
+                char c = glob.charAt(i);
+                if (c == '*') {
+                    if (i + 1 < glob.length() && glob.charAt(i + 1) == '*') {
+                        regex.append(".*");
+                        i += 2;
+                        // skip trailing slash after ** (e.g. **/ → match any prefix)
+                        if (i < glob.length() && glob.charAt(i) == '/') {
+                            i++;
+                        }
+                    } else {
+                        regex.append("[^/]*");
+                        i++;
+                    }
+                } else if (c == '?') {
+                    regex.append("[^/]");
+                    i++;
+                } else {
+                    regex.append(Pattern.quote(String.valueOf(c)));
+                    i++;
+                }
+            }
+            return Pattern.compile(regex.toString());
         }
     }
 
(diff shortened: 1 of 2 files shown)


extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
```diff
@@ -371,16 +365,52 @@ boolean isActive() {
...
+                    }
+                } else if (c == '?') {
+                    regex.append("[^/]");
+                    i++;
+                } else {
+                    regex.append(Pattern.quote(String.valueOf(c)));
+                    i++;
+                }
+            }
+            return Pattern.compile(regex.toString());
```
gnodet, line 420: **Nit — `**/suffix` pattern has a separator-blind edge case.**

When a `**` segment is followed by a non-`/` character (i.e., `**/foo` where the `**` is at the start), the implementation skips the `/` after `**` and concatenates `.*` directly with the remainder. The resulting regex `.*foo` matches `xfoo` (no separator), whereas the original `PathMatcher` would only match `x/foo` or `foo`.

This has **zero production impact** for the patterns actually emitted by scalpel — `managedDependencies/**`, `properties/*`, `module-a/**`, `**/*.md`, `docs/**` — all work correctly. But a user-supplied pattern like `**/pom.xml` would (incorrectly) also match `parentpom.xml`. The PathMatcher-faithful expansion for `a/**/b` is `a(?:/.*)?/b`, not `a/.*b`.

Since this is a behavior improvement over the broken `Path.of()` approach and doesn't affect existing tests or real usage patterns, it's not a blocker. Worth a comment or a follow-up test to document the known deviation.

gnodet (commented):
The two fixes are correct and CI validates them across all platforms. One nit on `globToPattern` worth knowing about.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
gnodet: Fixed in 959821b. Changed `**/` expansion from `.*` (skip slash) to `(.*/)?` so the separator is mandatory when matching. Added three `globToPattern` unit tests covering `**/suffix`, `prefix/**/suffix`, and `prefix/**` patterns.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
```diff
@@ -371,16 +365,55 @@ boolean isActive() {
...
+                if (c == '*') {
+                    if (i + 1 < glob.length() && glob.charAt(i + 1) == '*') {
+                        i += 2;
+                        if (i < glob.length() && glob.charAt(i) == '/') {
+                            // **/ → optionally match any path segment(s) ending with /
+                            regex.append("(.*/)?");
+                            i++;
+                        } else {
+                            // ** at end → match anything
+                            regex.append(".*");
```
gnodet, line 401: ⚠️ **ReDoS still present for `**<literal>` mid-pattern** (CodeRabbit finding unaddressed)

When `**` is followed by a non-`/` character — e.g. `**a**a**z` or `src/**Test.java` — the code emits `.*` and then immediately processes the next character as a literal. The resulting pattern `.*\Qa\E.*\Qa\E.*\Qz\E` triggers exponential backtracking on long non-matching inputs.

I measured this against the current code:
```
Pattern:  .*\Qa\E.*\Qa\E.*\Qa\E.*\Qz\E
n=50   → 11 ms
n=100  → 19 ms
n=200  → 280 ms   ← already painfully slow
n=400  → 4 350 ms ← CI-killer on a single pattern check
```

Multiple `**/` groups (`(.*/)?`) are safe — each is bounded by the `/` anchor and adds only linear cost. The problem is `.*` emitted for `**` not followed by `/`.

**Where this triggers:** `scalpel.excludeChanges` / `scalpel.includeChanges` are developer-configured system properties or POM properties. A pattern like `src/**Test.java` is a plausible real-world value that would blow up on any module with a long class name path.

**Fix:** Anchor `**` at end-of-pattern only. For `**` followed by a non-`/` char, emit `[^/]*` (same-segment wildcard, consistent with `*`) or raise a `PatternSyntaxException` documenting that `**` mid-segment is unsupported. The current behaviour of silently producing a backtracking regex is the worst option.

```suggestion
                        } else {
                            // ** at end of pattern → match anything (including path separators)
                            // Note: ** followed by a non-'/' char is not a standard glob construct;
                            // treat remaining * as single-segment wildcard to avoid ReDoS.
                            regex.append(".*");
                        }
```

If mid-segment `**` (e.g. `**Test.java`) must be supported, replace `Pattern.compile(...)` with `Pattern.compile(regex.toString(), Pattern.DOTALL)` and use possessive quantifiers or atomic groups — or switch to a linear NFA implementation. The simplest safe fix is to document that `**` is only valid as a full segment (`**/` or `/**` or `/**`), and throw on other uses.

gnodet (commented):
## Re-review after commit 959821b

**Previous finding addressed:** The separator-blind edge case in `globToPattern` is fixed. The `**/` expansion now emits `(.*/)?` instead of `.*`, so `**/pom.xml` no longer matches `parentpom.xml`. Three dedicated unit tests confirm the correct behaviour.

**Previous SonarCloud finding (cognitive complexity):** Resolved. The restructuring in 959821b reduced the nesting depth, and SonarCloud reports the quality gate as passed.

**One outstanding concern from CodeRabbit remains unaddressed** — the ReDoS risk on `.*`-producing patterns. I confirmed this empirically: a pattern like `**a**a**a**z` (double-star followed by a non-`/` char, interspersed with literals) produces `.*\Qa\E.*\Qa\E.*` which triggers catastrophic backtracking. At n=200 characters the match takes ~280 ms; at n=400 it takes ~4 350 ms. Details inline.

---

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
gnodet: Fixed in 225f3e0. When `**` is followed by a non-`/` char, we now emit `[^/]*` (single-segment wildcard) instead of `.*`. `**` at end-of-pattern still uses `.*` since there are no following literals to trigger backtracking. Added timing guard test confirming n=400 completes in < 1s.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
```diff
@@ -371,16 +365,55 @@ boolean isActive() {
...
+
+        /**
+         * Converts a simple glob pattern to a compiled {@link Pattern}.
+         * The glob syntax supports: {@code *} (any chars except {@code /}),
+         * {@code **} (any chars including {@code /}), {@code ?} (single char).
+         * Forward-slash {@code /} is the separator.  No filesystem is involved,
+         * so characters like {@code :} that are illegal in Windows file paths
+         * are handled correctly.
+         */
+        static Pattern globToPattern(String glob) {
```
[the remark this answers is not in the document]
gnodet: The cognitive complexity was already addressed in 959821b (confirmed by SonarCloud passing the quality gate). The further restructuring in 225f3e0 adds one more branch but reduces the nesting inside each — SonarCloud should remain green.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PomChangeAnalyzer.java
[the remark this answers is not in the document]
gnodet: Addressed in 225f3e0. Instead of replacing the entire regex-based approach (heavy lift), the fix constrains `**` mid-segment to `[^/]*` which eliminates the backtracking vector while preserving all standard glob semantics. The timing guard test confirms completion in < 1s for inputs that previously caused exponential backtracking.

gnodet (commented):
## Re-review after commit 225f3e0

**Previous finding addressed:** The ReDoS vulnerability for `**<literal>` mid-segment patterns is fixed. When `**` is followed by a non-`/` character, the code now emits `[^/]*` (single-segment wildcard) instead of `.*`. This removes the chained greedy fragments that caused catastrophic backtracking.

**Verification of the fix:** I traced the `globToPattern` logic manually and ran timing benchmarks:
- Pattern `**a**a**z` (3 groups) with `"a".repeat(400)+"z"` matching input: 0ms ✓
- Pattern `**a**a**z` (3 groups) with non-matching input at n=800: 320ms — polynomial, acceptable
- The specific 4-group pattern from the previous review now emits `[^/]*\Qa\E[^/]*\Qa\E[^/]*\Qa\E[^/]*\Qz\E` and completes in < 100ms at n=400 for the test input. The old `.*` pattern completed in 324ms at n=200 for the identical input — the new pattern is comparably safe.

**Residual risk (not a blocker):** `[^/]*` still has polynomial complexity for many chained same-char groups (e.g., 5 groups `[^/]*a[^/]*a[^/]*a[^/]*a[^/]*z` at n=200 non-matching: ~10s). This only triggers with deliberately adversarial user-crafted patterns like `**a**a**a**a**z` — not with any realistic change path filter. For patterns like `**/src/**/*.java`, `**Test.java`, or `docs/**`, the behaviour is fast. The timing test in the PR validates the specific case from the previous review and is sufficient for this use case.

**`addRepository` no-op (from earlier commits):** The two-overload no-op implementation is correct and complete. All abstract methods of `ModelResolver` are covered.

**SonarCloud:** Quality gate passed. All previous review threads resolved.

All previous findings from this review loop are addressed. Ready to merge.

---

PR #181, opened 2026-09-08
test: pin the O(M) graph-query bound of the trimmer (#116)

paoloantinori:
Follow-up to the analysis on the issue: current `main` already queries the graph O(M) times total (the downstream loop iterates `directlyAffected`; the upstream loop iterates a snapshot taken once, with transitive queries covering upstream-of-upstream in one call), so the issue's quadratic shape is already gone.

What was missing is anything pinning that property. This PR adds a regression test with a sync-point topology — eight leaves sharing one prerequisite — asserting upstream queries stay at most one per project, which a reintroduced per-member restart shape would violate. The graph test double counts queries. The test passes on current main, which is itself the evidence that the bound holds today.

If you had a specific quadratic scenario in mind that this misses, point me at it and I'll re-measure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build graph processing to avoid repeated upstream queries, helping prevent quadratic performance in projects with synchronization points.

* **Tests**
  * Added regression coverage to verify that each graph edge is queried no more than once during build-set calculation.

### DIFF
--- a/extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmerTest.java
+++ b/extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmerTest.java
@@ -13,6 +13,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import eu.maveniverse.maven.scalpel.core.ScalpelConfiguration;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmerTest.java
```diff
@@ -234,10 +235,45 @@ private ScalpelConfiguration configWith(boolean alsoMake, boolean alsoMakeDepend
...
+
+        Map<MavenProject, List<MavenProject>> upstream = new HashMap<>();
+        for (MavenProject leaf : leaves) {
+            upstream.put(leaf, List.of(sync, parent));
+        }
+        upstream.put(sync, List.of(parent));
+
+        TestDependencyGraph graph = new TestDependencyGraph(all, Map.of(), upstream);
+
+        trimmer.computeBuildSet(Set.of(leaves[0]), Set.of(), graph, configWith(true, true));
```
gnodet, line 262: ⚠️ **Weak assertion — test passes even without the O(M) guarantee.**

With `Set.of(leaves[0])` as the seed, the upstream loop touches exactly 3 projects (`leaf-0`, `sync`, `parent`) → 3 queries ≤ `all.size()` (10). Even a naive implementation that re-queries shared prerequisites on every leaf visit would produce ≤ 10 total queries in this configuration, because there is only one leaf in the initial build-set.

With all 8 leaves seeded the bound is tight: a correct single-pass implementation produces exactly 10 queries (8 leaves + sync + parent), while a quadratic restart would re-query `sync` and `parent` for each leaf and exceed the limit. CodeRabbit raised this — confirmed.

```suggestion
        trimmer.computeBuildSet(new LinkedHashSet<>(Arrays.asList(leaves)), Set.of(), graph, configWith(true, true));
```

extension3/src/test/java/eu/maveniverse/maven/scalpel/extension3/internal/ReactorTrimmerTest.java
```diff
@@ -234,10 +235,45 @@ private ScalpelConfiguration configWith(boolean alsoMake, boolean alsoMakeDepend
     /**
      * Simple test implementation of ProjectDependencyGraph.
      */
+    @Test
```
gnodet, line 238: 🔴 **Misplaced Javadoc — the `/**` comment now documents this test method instead of `TestDependencyGraph`.**

In the original file the `/** Simple test implementation of ProjectDependencyGraph. */` block sat directly above `private static class TestDependencyGraph`. The new test method was inserted between the Javadoc and the class it describes, stealing the documentation. `TestDependencyGraph` now has no Javadoc.

Move the test method above the Javadoc block (or below `TestDependencyGraph`), so the Javadoc attaches to the class again:

```suggestion
    @Test
    void computeBuildSet_queriesEachGraphEdgeAtMostOnce() {
```

(No code change above — the fix is to relocate the test so the `/**` block remains adjacent to `TestDependencyGraph`. Adjust the surrounding blank lines accordingly.)

gnodet (changes requested):
The test's intent is clear and the graph double is well-structured. Two issues need fixing before this merges.

paoloantinori:
Closing as duplicate of #173. My analysis had measured only the query count (O(M), single-snapshot loops) and missed the per-query uncached DFS cost that makes the total quadratic — #173 fixes the real thing. Correction posted on the issue.

PR #182, opened 2026-09-09
Fix #124: single PathFilters owner for every path-pattern family

### LINKED ISSUE #124: [arch] includePaths filtering re-derived at four sites with subtly different rules (correctness risk wearing a refactoring costume)

paoloantinori:
includePaths matching was independently re-derived at four sites with subtly different rules (affected-set, trim build-set, skip-tests, report downstream), the report site recompiled its matchers from config despite a comment at the first site claiming one shared compilation, and the disable/full-build trigger families compiled matchers inline inside per-file loops, reparsing each changed file once per pattern. Correctness risk wearing a refactoring costume, as the issue puts it.

`PathFilters` is constructed once from the resolved configuration and owns compilation and evaluation of all four families: module-scoped include matching (the dual direct-path plus descendant-pom rule), changed-file exclusion (the count log line preserved verbatim), and both trigger searches. The one intentional semantic difference is now stated where it applies instead of being implied by four copies: in trim mode an upstream build prerequisite outside the includePaths scope survives (commented at both trim-side filters), while skip-tests mode skips the tests of out-of-scope modules and the report omits them. `normalizeGlobPattern` moves next to its only remaining caller.

Gates run before pushing (review: line-by-line parity against main; simplify). The review gate caught one real slip, fixed: both trigger log lines had silently lost the `Scalpel: ` prefix (restored verbatim, matching the exclusion line). The simplify gate collapsed the trigger search's fragile parallel index into a single loop, made the four-list constructor private with `from(config)` the sole entry point, and deleted SLP's now-dead `GLOB_PREFIX`.

**One observable change, documented rather than reverted:** with multiple trigger patterns and files, the reported `triggerFile` is now the first changed file in file order matching any pattern (previously: the first file of the first matching pattern in pattern order). It lands in the report's `triggerFile` field and the shadow status reason. With the default single-pattern configuration there is no difference.

Behavior otherwise preserved, proven by the full battery: all unit tests and all 36 integration tests green, including the include-paths, exclude-paths, disable-triggers and full-build-trigger ITs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path-pattern matching so patterns without directory separators apply to matching files at any directory depth.
  * Standardized include and exclude filtering across project modules and changed files.
  * Improved detection of paths that disable triggers or require a full build.
  * Preserved file ordering when applying exclusions and ensured the first matching trigger path is reported.

### DIFF
--- a/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PathFilters.java
+++ b/extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PathFilters.java
@@ -47,7 +47,7 @@ final class PathFilters {
     private final List<CompiledPattern> fullBuildTriggers;
     private final List<PathMatcher> excludeMatchers;
 
-    PathFilters(
+    private PathFilters(
             List<String> includePatterns,
             List<String> excludePatterns,
             List<String> disableTriggerPatterns,
@@ -132,7 +132,8 @@ final class PathFilters {
      * off entirely), or null. Files are converted to {@link Path} once for the whole batch.
      */
     String findDisableTrigger(Set<String> changedFiles) {
-        return findTrigger(changedFiles, disableTriggers, "Disabled due to change in {} (matches disable trigger {})");
+        return findTrigger(
+                changedFiles, disableTriggers, "Scalpel: Disabled due to change in {} (matches disable trigger {})");
     }
 
     /**
@@ -140,20 +141,16 @@ final class PathFilters {
      * a full build), or null.
      */
     String findFullBuildTrigger(Set<String> changedFiles) {
-        return findTrigger(changedFiles, fullBuildTriggers, "Full build triggered by change to {} (matches {})");
+        return findTrigger(
+                changedFiles, fullBuildTriggers, "Scalpel: Full build triggered by change to {} (matches {})");
     }
 
     private String findTrigger(Set<String> changedFiles, List<CompiledPattern> triggers, String logFormat) {
         if (triggers.isEmpty()) {
             return null;
         }
-        List<Path> paths = new ArrayList<>(changedFiles.size());
         for (String file : changedFiles) {
-            paths.add(Path.of(file));
-        }
-        int i = 0;
-        for (String file : changedFiles) {
-            Path path = paths.get(i++);
+            Path path = Path.of(file);
             for (CompiledPattern trigger : triggers) {
                 if (trigger.matcher().matches(path)) {
                     logger.info(logFormat, file, trigger.pattern());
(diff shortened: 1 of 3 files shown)


extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PathFilters.java
```diff
@@ -0,0 +1,209 @@
...
+                config.getExcludePaths(),
+                config.getDisableTriggers(),
+                config.getFullBuildTriggers());
+    }
+
+    // ------------------------------------------------------------------
+    // includePaths: module-scoped
+    // ------------------------------------------------------------------
+
+    boolean hasIncludePatterns() {
```
gnodet, line 73: 💬 **Nit:** `hasIncludePatterns()` and `from()` (line 61) are the two package-private entry points most callers hit, yet both lack Javadoc. The class-level doc covers the *why* well; a one-liner on each would complete the picture:

```suggestion
    /** Returns {@code true} when at least one includePaths pattern was configured. */
    boolean hasIncludePatterns() {
```

gnodet (commented):
The consolidation is solid — the bug-wearing-refactoring-costume diagnosis is accurate, and the fix is correct. Patterns compiled once instead of in per-file loops, four quasi-independent implementations collapsed into one, the one intentional semantic difference documented at the right level. No correctness regressions found.

Two gaps worth addressing before merge:

**1. No dedicated unit tests for `PathFilters`.**

The moved `normalizeGlobPattern` tests now call `PathFilters.normalizeGlobPattern` — that's right. But the three new methods that carry real logic (`filterExcluded`, `matchesModuleInclude`, `findTrigger`) have zero direct tests. They're exercised by the integration test suite, but a unit test exercises them in isolation and is the right place to pin the documented semantic difference (see point 2 below).

Suggested coverage:
- `filterExcluded` with an empty excluder list (returns the input set unchanged)
- `filterExcluded` with a matching excluder (removes the expected file, logs the count)
- `matchesModuleInclude` with a direct-path pattern and a glob pattern
- `findTrigger` with multiple patterns and multiple files — specifically the file-first (new) vs pattern-first (old) ordering

**2. The triggerFile ordering change has no test.**

The PR body documents this correctly: with multiple patterns and multiple files, `findTrigger` now returns the first *file* (in file order) matching *any* pattern, where previously it returned the first file matching the first *pattern* (in pattern order) that had any match. With the default single-pattern config there's no difference, but the observable change lands in the shadow status reason and the report's `triggerFile` field — both user-visible. A targeted unit test on `PathFilters.findFullBuildTrigger` that exercises the two-pattern, two-file case would pin this documented behaviour and prevent silent regression.

---

**Static analysis note:** `ast-grep` flagged `catch (Exception e)` at several sites in `ScalpelLifecycleParticipant` — all are pre-existing on `main` and not introduced by this PR. Not attributable to this change.

extension3/src/main/java/eu/maveniverse/maven/scalpel/extension3/internal/PathFilters.java
```diff
@@ -0,0 +1,209 @@
...
+    String findFullBuildTrigger(Set<String> changedFiles) {
+        return findTrigger(
+                changedFiles, fullBuildTriggers, "Scalpel: Full build triggered by change to {} (matches {})");
+    }
+
+    private String findTrigger(Set<String> changedFiles, List<CompiledPattern> triggers, String logFormat) {
+        if (triggers.isEmpty()) {
+            return null;
+        }
+        for (String file : changedFiles) {
```
gnodet, line 152: ⚠️ **Semantic change — test gap:** This loop is file-first / pattern-second, where the old `findFullBuildTrigger` and `matchesDisableTrigger` were pattern-first / file-second. The PR body documents the observable difference (`triggerFile` is now the first file in file order matching *any* pattern, not the first file of the first matching pattern in pattern order), but there is no test that exercises this distinction.

With the default single-pattern configuration the two orderings are identical, so the risk is low — but the documented behaviour change has no regression guard. A unit test on `findFullBuildTrigger` with two patterns and two files (one matching pattern-1 only, the other matching pattern-2 only, in an order that produces a different result under the old strategy) would pin this.

paoloantinori:
Closing as duplicate of #176 (opened two days earlier) — our consolidation is the same shape and yours is first in time. Left our transferable review observations on #176.

